PR #5, opened 2025-10-15
Adding DPD workflow to topology builders

StephMcCallum:
## Description
The purpose of this code is to add a soft repulsion simulation, using DPD in HOOMD-blue as a building method for MuPT.

## Todos
  - [x] Integrate generators
  - [x] Call random walk generator directly in this python file instead of using two builders
  - [x] Add type hints
  - [x] Move functions into a DPD class
  - [x] Restructure DPD workflow to get information from mupt primitive
  - [x] Include calculation of translation and orientation transforms from final bead placement

## Status
- [x] Passes CI tests
- [x] Examples are updated and validated
- [x] Ready to go

### DIFF
--- a/mupt/builders/dpd.py
+++ b/mupt/builders/dpd.py
@@ -341,7 +341,6 @@ class DPDRandomWalk(PlacementGenerator):
             if (total_steps_run >= self.n_steps_max):
                 LOGGER.warning(f'Some particles are still too close after maximum simulation step {self.n_steps_max} reached; terminating simulation early')
                 break
-        gsd1.flush()
         end_time = time.perf_counter()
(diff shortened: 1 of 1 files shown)


devtools/conda-envs/dev-env.yml
timbernat: Worth revisitng whether these version pins also need to be applied to the release env

mupt/builders/base.py
timbernat: Addition of single-line of whitespace; might be stray?

timbernat (approved):
Achieves minimum-working DPD build, will pull to keep ball rolling

Issue #11, opened 2025-10-29
Canonical forms and serialization for Primitives


timbernat:
Currently, [Primitive](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L74) does not afford a native mechanism for being stored/reloaded from files; further, label components of PrimitiveHandles are currently user-assigned (or defaulted) string, which are allowed to be arbitrary and are not necessarily semantically meaningful. Support for Primitive serialization and canonical forms would greatly improve sharability of molecular systems built with or imported into MuPT, and further [TRUE](https://www.tandfonline.com/doi/full/10.1080/00268976.2020.1742938) software design. Further, canonical forms for Primitives based on their structural and information content would facilitate vastly-faster isomorphism and automorphism identification in hierarchically-organized complex molecular systems.

## Planned features
* Serialization to JSON/dict would be a natural first step, though some care is needed
* Relational database-like serialization and operations (e.g. [SQL](https://docs.python.org/3/library/sqlite3.html)) would be useful, particularly for managing internal/external Connector registries, though the additional complexity this would add may not be worth it to refactor to
* Canonical forms for each of the core MuPT Components ([Primitive](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L74), [Connector](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/connection.py#L127), [BoundedShape](https://github.com/MuPT-hub/mupt/blob/main/mupt/geometry/shapes.py#L77), and [TopologicalStructure](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/topology.py#L26)) could potentially be used to serialize hierarchy data compactly, depending on how dependent these canonical forms are on hashes or whether those forms are invertible
  * For example, TopologicalStructure can be serialized to [graphviz](https://graphviz.readthedocs.io/en/stable/) or [DOT](https://en.wikipedia.org/wiki/DOT_(graph_description_language)) or similar, which might be included as a field in a JSON-like format; care will need to be taken here for deeply-nested hierarchies (though a lot of potential for storage compression lies there as well) 

## Prior work
[...]

PR #22, opened 2026-01-05
Hierarchical Exporters: RDKit, OpenFF, and MDAnalysis with residue/chain metadata

joelaforet:
# Hierarchical Export Pipeline for MD Workflows

## Problem / Point of Friction

**MuPT's existing [primitive_to_rdkit()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/rdkit/exporters.py#L53) function flattens the Primitive hierarchy before export, permanently discarding residue and chain metadata.** This creates a critical blocker for molecular dynamics workflows that require hierarchical organization.

When building polymer systems in MuPT, we naturally construct a hierarchy: Universe → Chains → Residues → Atoms. This hierarchy encodes chemically meaningful information: which atoms belong to which monomer, which monomers form a chain, which chains comprise the system. However, the original exporter recursively flattens all Primitives into a single list of atoms, losing the grouping information entirely.

### Impact on Downstream Tools

**OpenMM requires residue and chain identification** when constructing a Topology. From the [OpenFF documentation](https://docs.openforcefield.org/projects/toolkit/en/latest/users/molecule_conversion.html):

> "OpenMM _requires_ identification of chains and residues when constructing a Topology, and residues must be entirely contained within their parent chain."

Without this metadata, [Topology.to_openmm()](https://github.com/openforcefield/openff-toolkit/blob/main/openff/toolkit/topology/topology.py#L1778-L1936) falls back to default values (`residue_name="UNK"`, `residue_number="0"`, `chain_id="X"`), producing a topology where every atom appears to belong to the same residue in the same chain.

**OpenFF Molecules support [hierarchy schemes](https://github.com/openforcefield/openff-toolkit/blob/main/openff/toolkit/topology/molecule.py#L1406-L1600)** that enable iteration via `mol.residues` and `mol.chains`. These read atom metadata fields (residue_name, residue_number, chain_id, insertion_code) populated from RDKit's [AtomPDBResidueInfo](https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.html#rdkit.Chem.rdchem.AtomPDBResidueInfo) during conversion. Without these fields set, calls to `mol.residues` return empty iterators—you cannot iterate over monomers in your polymer.

**MDAnalysis [selection language](https://docs.mdanalysis.org/stable/documentation_pages/selections.html)** relies on residue/chain metadata for queries like `select_atoms("resname HEA or resname TAL")` or `select_atoms("segid A")`. Without this information, per-residue RMSF calculations, residue-residue contact analysis, inter-chain radial distribution functions, and monomer-specific property calculations become impossible.

**Before this PR:** Users could build hierarchical polymer systems in MuPT, but exporting them destroyed the hierarchy, forcing manual reconstruction of residue/chain information—a tedious, error-prone process.

**After this PR:** The full hierarchy flows through the export pipeline: MuPT → RDKit (via `AtomPDBResidueInfo`) → OpenFF (via atom metadata + hierarchy schemes) → MDAnalysis (via `atom_resindex` and `residue_segindex` mappings).

### Terminology

I define the hierarchical schema as:

- **[Universe]** – Root node containing all distinct molecules as children (the "univprim")
- **[Chains]** – Primitives representing distinct molecular fragments (non-covalently bonded entities). A box with two 5-monomer polymers has 2 chainz
- **[Residues]** – Primitives representing repeated atom groupings (monomers)
- **[Atoms]** – Leaf Primitives representing individual atoms

This satisfies the contracts of [building an MDAnalysis Universe from scratch](https://docs.mdanalysis.org/dev/documentation_pages/core/universe.html#MDAnalysis.core.universe.Universe.empty), which requires:
- n_atoms ([int](https://docs.python.org/3/library/functions.html#int)) – number of Atoms in the Universe
- n_residues ([int](https://docs.python.org/3/library/functions.html#int)), default 1) – number of Residues in the Universe
- n_segments ([int](https://docs.python.org/3/library/functions.html#int), default 1) – number of Segments in the Universe (maps to our "Chains")
- atom_resindex (array like) – mapping of atoms to residues
- residue_segindex (array like) – mapping of residues to segments

Note, there are other parameters that go into the construction of the Universe, but the above are the most important.

### Three-Layer Export Pipeline

Rather than a monolithic exporter, we built a layered system that allows interception at any stage:

```
Primitive (MuPT) 
    └─→ RDKit Mol (with AtomPDBResidueInfo) 
            └─→ OpenFF Molecule (with hierarchy schemes)
                    └─→ MDAnalysis Universe
```

Users can stop at:
- **RDKit layer** – For cheminformatics workflows, substructure searches, 2D depiction
- **OpenFF layer** – For force field parameterization via [Interchange](https://docs.openforcefield.org/projects/interchange/en/stable/)
- **MDAnalysis layer** – For trajectory analysis and file I/O

This design provides multiple insertion points for future export targets. For example, when we eventually build interfaces to HOOMD and Foyer.

### Key Features

**Bond order preservation:** Both intra-residue and inter-residue bonds are extracted with their bond orders (single, double, aromatic, etc.) from MuPT's `Connector.bondtype` attribute. This ensures correct aromaticity perception in RDKit, accurate force field typing in OpenFF, and proper bond order attributes in MDAnalysis. See the implementation [here](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/rdkit/exporters.py#L316-L351) and [here](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/rdkit/exporters.py#L353-L412).

**PDB-compliant naming:** We enforce 3-character uppercase residue names via the `resname_map` parameter and [_pdb_resname()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/mdanalysis/exporters.py#L22) helper, ensuring compatibility with PDB format specifications and visualization tools like VMD and PyMOL.

**Chain ID generation:** Chain IDs are auto-generated as single characters (A, B, C, ...) with overflow handling (AA, AB, ...) for systems with >26 chains, following [PDB format conventions](https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html).
NOTE: If we plan to support solvent molecules in the MuPT representation + system export, we may want to group all water molecules into the same "Chain", even though they are not covalently linked.

**Flexible coordinate handling:** Atom positions are extracted from `atom.shape.centroid` when available, falling back to [0, 0, 0] for Primitives without geometry. This allows the same exporter to work for both coordinate-containing and coordinate-free representations.

### Completed
- [x] [primitive_to_rdkit_hierarchical()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/rdkit/exporters.py#L150) – RDKit export preserving chain/residue metadata via `AtomPDBResidueInfo`
- [x] [primitive_to_openff_molecules()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/openff/exporters.py#L40) – OpenFF Molecule export with hierarchy schemes for residue iteration
- [x] [primitive_to_openff_topology()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/openff/exporters.py#L137) – OpenFF Topology export for multi-chain systems
- [x] [primitive_to_mdanalysis()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/mdanalysis/exporters.py#L79) – MDAnalysis Universe export with full topology attributes
- [x] [save_openff_system()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/openff/serialization.py#L24) / [load_openff_system()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/openff/serialization.py#L106)– JSON + SDF serialization for OpenFF molecules
- [x] [build_copolymer_system()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/builders/copolymer.py#L148)– Random copolymer generation builder with proper hierarchy
- [x] Bond order preservation throughout the export pipeline
- [x] Validation via [_is_AA_export_compliant()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/mdanalysis/exporters.py#L68) to enforce universe → chain → residue → atom hierarchy

### Remaining Work
- [ ] Comprehensive unit tests for roundtrip preservation of residue/chain metadata
- [ ] Tutorial notebook demonstrating full workflow from MuPT construction to MD simulation
- [ ] Consider supporting non-AA-compliant hierarchies (arbitrary depth) in future iterations

## Open Questions

- How should we support export of coarse-grained Primitives (non-atomic leaves) to these formats?
- Should [_pdb_resname()](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/mdanalysis/exporters.py#L22) raise an error or silently truncate for labels longer than 3 characters?
- What's the best approach for interfacing with the project's existing testing infrastructure?

## Notes

- The polymer build code was abstracted from [ellipsoidal_chain_placement.ipynb](https://github.com/MuPT-hub/mupt-examples/blob/main/examples_system/ellipsoidal_chain_placement.ipynb)—when the DPD placer code is finished, we should integrate it into [copolymer.py](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/builders/copolymer.py)
- Bond order + Bonding Information extraction was challenging to implement; additional documentation and clearer explanation of the `Connector` concept would be valuable
- My choice of naming convention for the fields in Terminology is based on what protein folks commonly call these things. We can change the names of them to be more general if needed. I.E. Atom == Particle, Chain == Molecule, etc.

## Status: 
- [ ] Ready for review

---


mrshirts:
Hi, @joelaforet 

This seems like it will work well to generate some MD files using the representation to test things out.  I don't think we want to merge this branch in, because we don't want to have dependencies on OpenFF and even RDKit - we want to copy the information over to the MD Analysis Universe directly from the representation.  It won't generalize that well to different resolutions.  

I think the hierarchical schema looks pretty close to what we want.  We will want some flexibility for what parts of the representation get assigned to which parts of the hierarchy, which will change - for example, 1 bead - 1 monomer models will have residues and atoms both at the same level, whereas all atom obviously won't, but we can start with the aa pathway for now. 

> PDB-compliant naming: 

We probably want to have more control over this - some applications might want PDB compliant naming, some won't. This is worth some discussion of how to do right. 

>Chain ID generation: 

This is complicated, since it's optional in the PDB.  WE should probably focus more on what MDAnalysis needs for chain identification, as it can export to PDB. 

> Bond order preservation:

This is complicated. We want to avoid having to do this if we have to - this is worth an issue of how much of this information is needed.   WE want to avoid OpenFF, because we might export via Foyer.  RDKit, however, is probably how we are going to get to Foyer, so MAYBE that can be in.  What features of bond order does MDAnalysis use?

joelaforet:
Hi, @mrshirts 

> I don't think we want to merge this branch in, because we don't want to have dependencies on OpenFF and even RDKit - we want to copy the information over to the MD Analysis Universe directly from the representation. It won't generalize that well to different resolutions.

I implemented multiple different ways of exporting a primitive that defines an all-atom molecular system. Going from:
 `Primitive->RDKit->OpenFF->MDAnalysis` is just **ONE** of the pathways in this PR. 
There **is** a direct Primitive->MDAnalysis function [here](https://github.com/MuPT-hub/mupt/blob/MDAexport/mupt/interfaces/mdanalysis/exporters.py#L80C1-L101C76). Unfortunately, there is not currently a mechanism for exporting a MuPT Primitive to file, see [here](https://github.com/MuPT-hub/mupt/issues/11). Let's suppose a user is following our tutorial notebooks and are in an active Jupyter notebook session. If one wanted to use MuPT to instantiate their system, then they would need to perform the simulation in the same session in order for the metadata to be preserved. The current mechanism of "Instantiation -> flattening -> RDKit export -> OpenFF -> OpenMM" outlined in [this notebook](https://github.com/MuPT-hub/mupt-examples/blob/main/examples_system/mdfiles_with_openff.ipynb) discards the metadata necessary for analyzing the system. Even if we use a "Primitive -> MDAnalysis" [exporter](https://github.com/MuPT-hub/mupt/blob/MDAexport/mupt/interfaces/mdanalysis/exporters.py#L80C1-L101C76), users would need to run all of their simulations in the Jupyter session they built the system in, or have the "Build system with MuPT" step as the first part of their simulation script they send to a HPC. This also won't interface with MD engines that aren't compatible with being run in a Jupyter Notebook session like LAAMPS or GROMACS.

This is why I implemented the exporters to RDKit and OpenFF explicitly: we need a way to keep the metadata so that a user can analyze their trajectory!

joelaforet:
Content was extracted from this issue and implemented in the following PR's separately:
MDAnalysis: #50 
RDKit: #58 #59 #60

PR #24, opened 2026-01-23
Add MDAnalysis Export Interface for SAAMR-compliant Primitives

joelaforet:
# Add MDAnalysis Export Interface for SAAMR Primitives

---
## Summary

This PR adds the ability to export Primitives in Standard All-Atom Molecular Representation (SAAMR) format to MDAnalysis `Universe` objects. This enables users to leverage MDAnalysis's powerful trajectory analysis capabilities on MuPT-generated molecular structures.

This is the MDAnalysis component of PR #22, split into smaller, focused PRs as requested. This PR implements the MDAnalysis export capability with preliminary tests.

---
## Motivation

[MDAnalysis](https://www.mdanalysis.org/) is a widely-used Python library for analyzing molecular dynamics trajectories. By providing a direct export path from MuPT `Primitives` to MDAnalysis `Universes`, users can:
- Load and analyze MD trajectories of MuPT-generated systems
- Utilize MDAnalysis's extensive analysis toolkit (RMSD, RDF, contacts, etc.)
- Perform residue-specific selections and analyses using MDAnalysis's [selection language](https://docs.mdanalysis.org/stable/documentation_pages/selections.html)
- Integrate MuPT into existing MDAnalysis-based workflows

---
### Why This Matters

MDAnalysis's selection language relies on residue/chain metadata for queries like `select_atoms("resname HEA or resname TAL")` or `select_atoms("segid A")`. This metadata enables:
- Per-residue RMSF calculations
- Residue-residue contact analysis  
- Inter-chain radial distribution functions
- Monomer-specific property calculations

Without proper export to MDAnalysis with hierarchical information preserved, these common analysis tasks become impossible or require manual reconstruction of the molecular hierarchy.

---
### New Functionality
- **[mupt/interfaces/mdanalysis/exporters.py](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/interfaces/mdanalysis/exporters.py)**: Core export implementation
  - [primitive_to_mdanalysis()](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/interfaces/mdanalysis/exporters.py#L78): Convert SAAMR Primitives to MDAnalysis Universe
  - [_pdb_resname()](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/interfaces/mdanalysis/exporters.py#L29): Helper for PDB-compliant 3-character residue naming
- **[mupt/mutils/saamr.py](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/mutils/saamr.py)**: New home for utilities associated with SAAMR compliant systems.
  - [_is_SAAMR_compliant()](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/mutils/saamr.py#L9): Validation for Universe → Chain → Residue → Atom hierarchy

### Testing Infrastructure
- **[mupt/tests/conftest.py](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/tests/conftest.py)**: Centralized pytest fixtures
  - Reusable fixtures for common test molecules and Primitives
  - Reduces code duplication across test modules
  - Establishes patterns for future test development
  
- **[mupt/tests/interfaces/mdanalysis/test_exporters.py](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/tests/interfaces/mdanalysis/test_exporters.py)**: Preliminary test suite
  - Tests for successful export of various molecular structures
  - Validation of atomic properties (element symbols, positions, bonds)
  - Error handling for non-atomizable Primitives
  - Achieves **>90% code coverage** for the exporter module

---
### SAAMR Compliance

**Standard All-Atom Molecular Representation (SAAMR)** defines a specific hierarchical organization:

- **Universe** – Root container holding all distinct molecules in the system
- **Molecules** – Collections of covalently-bonded repeat units (e.g., a single polymer chain)
- **Repeat Units** – Atom groupings with standard names (e.g., monomers like "HEA" or "TAL")
- **Atoms** – Individual atomic constituents (leaf nodes)

This hierarchy maps naturally to MDAnalysis's organizational structure:
- **Universe** → `MDAnalysis.Universe`
- **Molecules** → Segments (via `residue_segindex`)
- **Repeat Units** → Residues (via `atom_resindex`)
- **Atoms** → Atoms

The [_is_SAAMR_compliant()](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/mutils/saamr.py#L9) function validates that a Primitive adheres to this four-level hierarchy before export, ensuring compatibility with downstream MD tools that expect this standard organization.

---

### Key Features

**Bond topology preservation:** Bonds are extracted from MuPT's `Connector` objects and properly added to the MDAnalysis Universe topology, ensuring accurate molecular structure representation.
- Implementation of Inter-Repeat_Unit bond extraction: [(here)](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/interfaces/mdanalysis/exporters.py#L235-L311)
- Implementation of Intra-Repeat_Unit bond extraction: [(here)](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/interfaces/mdanalysis/exporters.py#L161-L233)

Both of the above implementations rely on **expensive nested for-loops** to extract the necessary information. **This is prototype code that does its job**, however there are certainly better ways of traversing the Tree structure of the MuPT representation. This region should be priority number 1 for refactoring once we have a better sense of how we want people to extract bonds from a SAAMR (or non-SAAMR) Primitive. I.E. Depth-First Search instead of nested for loops.

**PDB-compliant naming:** Residue names are enforced as 3-character uppercase strings via the `_pdb_resname()` helper, ensuring compatibility with PDB format specifications and visualization tools like VMD and PyMOL. For now, it is the responsibility of the user to provide said naming schema as a dictionary. This is done at the time the user creates their system.

**Flexible coordinate handling:** Atom positions are extracted from `atom.shape.centroid` when available, falling back to [0, 0, 0] for Primitives without geometry. This allows the same exporter to work for multiple placement methods. I.E. Systems made with DPD, RandomWalk, etc.

---
## Testing

All tests pass using pytest with the following coverage:
```bash
pytest mupt/tests/interfaces/mdanalysis/ -v --cov=mupt/interfaces/mdanalysis
```

Tests validate:
- [ ] Atomic element preservation
- [ ] Positional data transfer
- [ ] Bond topology accuracy
- [ ] Proper handling of edge cases (single atoms, complex molecules)
- [ ] Appropriate error raising for invalid inputs
- [ ] Hierarchy validation via `_is_AA_export_compliant()`

## Dependencies

Adds `MDAnalysis` as an optional dependency for users who need trajectory analysis capabilities.

## Relationship to Original PR #22

This PR extracts the MDAnalysis export functionality from the larger PR #22. The original PR included:
- RDKit hierarchical export (`primitive_to_rdkit_hierarchical()`)
- OpenFF Molecule/Topology export
- MDAnalysis export (**this PR**)
- Copolymer builder
- Serialization utilities

Future PRs will add the remaining components with their own hierarchical metadata preservation capabilities.

**Breaking Changes**: None (to my knowledge)

**Related Issues**: Extracted from #22 (split into smaller PRs as requested)

---


mrshirts:
I'm not sure we should enforce the 3 character limit, as it limits what could be done.  I think this should be dealt with in export to PDB.

mupt/mutils/saamr.py
timbernat, line 1: This is essential to have in this PR but shouldn't live in `mutils`, which is intended for more boilerplate [stdlib](https://docs.python.org/3/library/index.html) functionality (e.g. reading files, copying objects, etc.). I'd have this module be `mupt.mupr.properties` or something similar, since we'll eventually have a whole host of Primitive properties used to determine exportability status; I have a [version such properties in another PR](https://github.com/MuPT-hub/mupt/blob/prim_refactor/mupt/mupr/primitives.py#L577-L627)

mupt/tests/interfaces/mdanalysis/test_exporters.py
```diff
@@ -0,0 +1,268 @@
...
+# ============================================================================
+# NEGATIVE TEST CASES: Verify proper error handling for invalid inputs
+# ============================================================================
+
+@pytest.mark.parametrize(
+    "primitive_builder,resname_map,expected_exception,test_id",
+    [
+        # Non-SAAMR compliant: Universe -> Atom directly (depth=1, should be 3)
+        (
+            lambda: (
```
timbernat: Some really wacky-looking lambdas in this section (not sure what these are meant to be doing?). Consider moving these into fixtures (which [can be referenced by name in `parametrize` blocks](https://stackoverflow.com/a/64348247/31240875)) or even just module-level constants (not ideal, but ok for here)

mupt/tests/interfaces/mdanalysis/test_exporters.py
```diff
@@ -0,0 +1,268 @@
...
+    expected_exception : Exception class
+        The exception type we expect to be raised
+    test_id : str
+        Human-readable test case identifier for debugging
+    """
+    # Arrange: Build the test primitive
+    univprim = primitive_builder()
+    
+    # Act & Assert: Verify the expected exception is raised
+    with pytest.raises(expected_exception):
```
timbernat: Sounds like a job for [pytest.mark.xfail](https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-xfail); you can use these in conjunction with parametrize statements ([example](https://github.com/timbernat/polymerist/blob/main/polymerist/tests/polymers/building/test_sequencing.py#L21-L54)) to avoid passing around Exceptions as arguments to tests

README.md
```diff
@@ -48,7 +48,7 @@ See the accompanying [examples repository](https://github.com/MuPT-hub/mupt-exam
 
 ### Copyright
 
-Copyright (c) 2024, Timotej Bernat
+Copyright (c) 2024, Timotej Bernat, Joseph R. Laforet Jr.
```
timbernat, line 51: This really ought to be everyone ("MuPT Contributors" or something to that effect)

mupt/tests/conftest.py
```diff
@@ -0,0 +1,618 @@
...
+                embed_positions=True,
+                label=unit_name,
+            )
+            
+            # Find edge atoms marked with atom map numbers
+            head_atom, tail_atom = unitprim.search_hierarchy_by(
+                lambda prim: 'molAtomMapNumber' in prim.metadata,
+                min_count=2,
+            )
+            head_pos = head_atom.shape.centroid
```
timbernat, line 118: This block from the examples is basically completely unnecessary for what you're testing here (sizes and assigns an Ellipsoid shape to each repeat unit and aligns coordinates to axes). Shouldn't affect MDA export tests if removed

mupt/tests/conftest.py
```diff
@@ -0,0 +1,618 @@
...
+    # Validate that required units are present
+    for required in [head_name, tail_name] + list(mid_distrib.keys()):
+        if required not in lexicon:
+            raise ValueError(f"Missing repeat unit '{required}' in rep_unit_smiles")
+    
+    # Build universe
+    univprim = Primitive(label='universe')
+    chain_lengths = np.random.randint(chain_len_min, chain_len_max + 1, size=n_chains)
+    
+    if show_progress:
```
timbernat: You don't need a progress bar for a unit test!

mupt/tests/conftest.py
```diff
@@ -0,0 +1,618 @@
...
+                nx.path_graph(
+                    molprim.children_by_handle.keys(),
+                    create_using=TopologicalStructure,
+                ),
+                max_registration_iter=100,
+            )
+            
+            # Place beads by random walk
+            direction = random_unit_vector()
+            # DEV-JRL: This is where we should think about inserting the DPD Builder!
```
timbernat, line 259: Eventually yes; we've had extensive discussion about this (and what the current limitations are) in the #representation channel on Slack.

mupt/tests/conftest.py
```diff
@@ -0,0 +1,618 @@
...
+        custom_ratio = BPA_BPS_factory(
+            chain_len_min=5,
+            chain_len_max=15,
+            n_chains=20,
+            bps_fraction=0.7  # 70% BPS, 30% BPA
+        )        
+'''
+
+@pytest.fixture
+def polyethane_smiles() -> dict[str, str]:
```
timbernat: Polyethane isn't a polymer; should rename this (and other occurences) to "polyethylene", as:
```suggestion
def polyethylene_smiles() -> dict[str, str]:
```

mupt/tests/conftest.py
```diff
@@ -0,0 +1,618 @@
...
+def polyethane_resname_map():
+    """Residue name mapping for polyethane systems."""
+    return {
+        'head': "HEA",
+        'ethane': "EAN",
+        'tail': "TYL"
+    }
+
+@pytest.fixture
+def BPA_BPS_smiles() -> dict[str, str]:
```
timbernat: (completely optional) the polymer system described here is [polyethersulfone (PES)](https://en.wikipedia.org/wiki/Polysulfone). Could rename to reflect that, but doesn't add much by way of clarity
```suggestion
def polyethersulfone_smiles() -> dict[str, str]:
```

mupt/tests/conftest.py
```diff
@@ -0,0 +1,618 @@
+'''
```
timbernat: (discussion) do these polymer building fixtures need to live in conftest, or are they better suited for definition in the test files that call them (or __init__ in a module of tests, etc.)?

mupt/builders/random_walk.py
```diff
@@ -1,7 +1,7 @@
 '''Utilities for generating coordinates of random walks'''
 
-__author__ = 'Timotej Bernat'
-__email__ = 'timotej.bernat@colorado.edu'
+__author__ = 'Timotej Bernat, Joseph R. Laforet Jr.'
```
timbernat: Not sure suppressing a deprecation warning needs an authorship change?

mupt/interfaces/mdanalysis/exporters.py
```diff
@@ -0,0 +1,379 @@
...
+from ...mutils.saamr import _is_SAAMR_compliant
+from ...chemistry.core import BOND_ORDER
+
+import logging
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.setLevel(logging.DEBUG)
+
+# DEV:JRL -> Debating whether to keep resname_map optional.
+def _pdb_resname(label: str, resname_map: Optional[dict]) -> str:
```
timbernat: Agree resname_map should be non-optional dict; if not provided, this functions is just a glorified length check
```suggestion
def _pdb_resname(label: str, resname_map: dict[str, str]) -> str:
```

If you do decide to keep it Optional, realize that in the code by providing a NoneType sentinel as default:
```suggestion
def _pdb_resname(label: str, resname_map: Optional[dict[str, str]]=None) -> str:
```

mupt/interfaces/mdanalysis/exporters.py
```diff
@@ -0,0 +1,379 @@
...
+        name = label
+
+    if len(name) != 3:
+        raise ValueError(
+            f"Residue name '{name}' (from '{label}') is not 3 characters long"
+        )
+    return name.upper()
+
+# DEV: JRL -> Debating whether to keep resname_map optional.
+# DEV: JRL -> I wrote this function with protein-flavored vocabulary in mind.
```
timbernat, line 77: (discussion) Thanks for lampshading this; I'm in favor of the less protein-centric names, but we should decide and rename vars accordingly before merging to avoid a painful compatibility rename later

mupt/interfaces/mdanalysis/exporters.py
```diff
@@ -0,0 +1,379 @@
...
+    
+    Examples
+    --------
+    >>> universe = primitive_to_mdanalysis(my_univprim)
+    >>> LOGGER.info(f"Created universe with {universe.atoms.n_atoms} atoms")
+    >>> LOGGER.info(f"Number of residues: {universe.residues.n_residues}")
+    >>> LOGGER.info(f"Number of segments: {universe.segments.n_segments}")
+    """
+    
+    assert _is_SAAMR_compliant(univprim), "[Under Construction] Primitive must be ordered according to: universe -> chains -> residues -> atoms"
```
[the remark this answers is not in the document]
timbernat: This should just be an Exception (maybe ValueError or even a custom one if you wanted to be more specific), since it's user facing rather than communicating with developers (which is what `assert` is for). Something like:
```suggestion
if not _is_SAAMR_compliant(univprim):
    raise ValueError('<message of choice>')
```

README.md
joelaforet: Agree, people should add themselves as they make PR's!

mupt/builders/random_walk.py
joelaforet: I think we should clarify as an org what the purpose of __author__ is. I was under the impression that this denotes people who have touched this file, and could potentially be blamed for it working/not working. That being said, I agree that my change doesn't put me on this list.
```suggestion
__author__ = 'Timotej Bernat'
```

mupt/builders/random_walk.py
```diff
@@ -1,7 +1,7 @@
...
+__email__ = 'timotej.bernat@colorado.edu, jola3134@colorado.edu'
```
joelaforet:
```suggestion
__email__ = 'timotej.bernat@colorado.edu
```

mupt/interfaces/mdanalysis/exporters.py
```diff
@@ -0,0 +1,379 @@
...
+from typing import Optional
+
+from ...mupr.primitives import Primitive
+from ...mutils.saamr import _is_SAAMR_compliant
+from ...chemistry.core import BOND_ORDER
+
+import logging
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.setLevel(logging.DEBUG)
```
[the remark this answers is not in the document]
joelaforet: Not sure what the right call is so I'm ignoring this suggestion.

mupt/builders/random_walk.py
```diff
@@ -1,7 +1,7 @@
 '''Utilities for generating coordinates of random walks'''
 
 __author__ = 'Timotej Bernat'
-__email__ = 'timotej.bernat@colorado.edu'
+__email__ = 'timotej.bernat@colorado.edu
```
joelaforet:
```suggestion
__email__ = 'timotej.bernat@colorado.edu'
```

mupt/interfaces/mdanalysis/exporters.py
joelaforet: (discussion) MDAnalysis calls them "chains/segID", "residues", and "atoms". I think we should stay consistent with each package's choice of nomenclature, but MuPT should remain "molecules", "repeat units", and "particles".

mupt/interfaces/mdanalysis/exporters.py
joelaforet: Agreed! It is required!

mupt/interfaces/mdanalysis/exporters.py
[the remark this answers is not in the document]
joelaforet: I did not realize that is the distinction for when one would use assert vs throw an error. Thank you!

mupt/interfaces/mdanalysis/exporters.py
```diff
@@ -0,0 +1,379 @@
...
+        # Add bond orders if available - must be passed during bonds creation
+        if bond_orders and len(bond_orders) == len(bonds):
+            # MDAnalysis Bonds class accepts order= parameter at creation time
+            bond_attr = Bonds(np.asarray(bonds, dtype=np.int32), order=bond_orders)
+            universe.add_TopologyAttr(bond_attr)
+            LOGGER.info(f"Added {len(bonds)} bonds with {len(bond_orders)} bond orders")
+            
+            # Show distribution of bond orders
+            from collections import Counter
+            order_counts = Counter(bond_orders)
```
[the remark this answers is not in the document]
joelaforet: Moved import! Resolved!

mupt/mutils/saamr.py
```diff
@@ -0,0 +1,18 @@
...
+__author__ = 'Joseph R. Laforet Jr.'
+__email__ = 'jola3134@colorado.edu'
+
+from ..mupr.primitives import Primitive
+
+def _is_SAAMR_compliant(prim : Primitive) -> bool:
+    """
+    Check whether a Primitive hierarchy is organized
+    as universe -> molecule -> repeat-unit -> atom
+    """
```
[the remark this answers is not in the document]
joelaforet: Resolved

mupt/mutils/saamr.py
joelaforet: That makes sense to me! I'll move this over to that file.

mupt/tests/conftest.py
joelaforet: (discussion) I don't know! I also don't have strong opinions on this issue, so I'll defer to you. I placed them here because I assumed that multiple tests will require building polymers, so best not to copy/paste the same polymer generation logic across multiple module test files.

mupt/tests/conftest.py
```diff
@@ -0,0 +1,618 @@
...
+
+# resname_maps are currently used to assign PDB compliant 3-char codes
+# to repeat units
+@pytest.fixture
+def polyethane_resname_map():
+    """Residue name mapping for polyethane systems."""
+    return {
+        'head': "HEA",
+        'ethane': "EAN",
+        'tail': "TYL"
```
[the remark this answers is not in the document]
joelaforet: Nope!

mupt/tests/conftest.py
joelaforet: Thanks for that context! I was thinking from a residue-first perspective, not seeing the whole polymer. I think that the name of the function should reflect the whole polymer, but the body should reflect that polymer's composition. I.E. I would not know how to map "polyethersulfone" to repeat units called "bisphenol_S" and "bisphenol_A", but it was clear to me that the "BPA/BPS copolymer" is composed of BPA and BPS units.

mupt/tests/conftest.py
joelaforet: Good point!

mupt/tests/conftest.py
joelaforet: Isn't this necessary for assigning positions to the Ellipsoids, though? I guess I'm confused what would happen if we tried to make ellipsoids that don't have coordinates and positions? I added this in anticipation of future tests that may rely on positions being assigned.
I.E. I'm imagining that when the DPD placer is implemented, we will need to test the Analysis codes by first building a system with MuPT, running a short simulation, computing the property, and then comparing to a known reference.

mupt/tests/conftest.py
```diff
@@ -0,0 +1,618 @@
...
+    Build a PSU/PES copolymer system:
+    
+    >>> rep_unit_smiles = {
+    ...     'head': '[H]-[O:1]c1ccc(cc1)S(=O)(=O)c1cc[c:2](cc1)-*',
+    ...     'bisphenol_S': '*-[O:1]c1ccc(cc1)S(=O)(=O)c1cc[c:2](cc1)-*',
+    ...     'bisphenol_A': '*-[O:1]c1ccc(cc1)C(-C)(-C)c1cc[c:2](cc1)-*',
+    ...     'tail': '*-[O:1]c1ccc(cc1)S(=O)(=O)c1ccc(cc1)[O:2]-[H]',
+    ... }
+    >>> mid_distrib = {'bisphenol_S': 0.4, 'bisphenol_A': 0.6}
+    >>> univprim = build_copolymer_system(
```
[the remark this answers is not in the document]
joelaforet: resolved!

mupt/tests/conftest.py
joelaforet: Doh!

mupt/tests/interfaces/mdanalysis/test_exporters.py
joelaforet: Full transparency, I used ChatGPT to find a way to test error messages. This was to get more coverage. At a high level, though, the lambdas serve as self-contained functions that return a univprim that I know is "wrong". The actual function `test_primitive_to_mdanalysis_error_handling` is an abstract method that takes in "some method of producing a primitive" and "some expected error" and determines if the expected error is thrown. I am open to suggestions on how to improve this!

README.md
mrshirts: We do need a overall MuPT policy here.  I think Joe's suggestion of adding themselves as they make a change is a good first step, I will put on the todo list for PIs tol come up with something more comprehensive soon.

mupt/tests/conftest.py
mrshirts: @timbernat could you comment here? Sounds like some discussion is needed to establish exactly what the code should carry out and the constraints.

mupt/tests/conftest.py
mrshirts: @timbernat can you follow up on the discussion to help resolve?

mupt/tests/interfaces/mdanalysis/test_exporters.py
joelaforet: The purpose of this test is to verify the correct exception is thrown. Xfail only checks if the test fails, not that a specific error is thrown, so I choose to keep pytest.raises().

mupt/tests/conftest.py
timbernat: I recommend this mostly to be consistent with [IUPAC's recommended conventions for polymer naming](https://iupac.org/wp-content/uploads/2019/07/140-Brief-Guide-to-Polymer-Nomenclature-Web-Final-d.pdf). Many small molecules have alternative names (e.g. [bisphenol S](https://cactus.nci.nih.gov/chemical/structure/c1cc(ccc1O)S(=O)(=O)c2ccc(cc2)O/names)), and in this case the ether bonds _only_ appear in the polymerized chain (from hydrolysis of -OHs in monomers). In general, it's not always easy to see where the common name for a polymer comes from just from looking at its repeat units.

mupt/tests/conftest.py
timbernat: > ellipsoids that don't have coordinates and positions?

I've designed Primitives such that applying a rigid transformation to one also applies that transformation [to all of its children](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L1094-L1095), even if the [parent Primitive does not itself have a shape](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L1087-L1088). 

What I'm saying is that if you remove lines [138-141](https://github.com/MuPT-hub/mupt/blob/c86e4e2e24dd20d91a18ede327a41625e7b381de/mupt/tests/conftest.py#L138-L141), the repeat unit-level Primitives will have `NoneType` shape, but the atom-level Primitives will still be assigned correct relative and absolute positions (via calls in [110](https://github.com/MuPT-hub/mupt/blob/c86e4e2e24dd20d91a18ede327a41625e7b381de/mupt/tests/conftest.py#L110) and [135](https://github.com/MuPT-hub/mupt/blob/c86e4e2e24dd20d91a18ede327a41625e7b381de/mupt/tests/conftest.py#L135), respectively), stored as single-element PointCloud instances in their `.shape` attr. This is where any exporter would access the positions of the atoms, e.g. the [RDKit exporter](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/interfaces/rdkit/exporters.py#L86) or the [MDAnalysis exporter](https://github.com/MuPT-hub/mupt/blob/MDAexport_clean/mupt/interfaces/mdanalysis/exporters.py#L191) here.

>  I'm imagining that when the DPD placer is implemented, we will need to test the Analysis codes by first building a system with MuPT, running a short simulation...

For DPD, or really any coarse grained simulation, the atom positions are not kept track of, only the repeat unit positions (and orientations, if we're lucky) are. From the above discussion, as long as the simulation tool hands the net transformation of each unit back to MuPT, they can be "backmapped" onto the atom positions below, which the MuPT representation _does_ store. The specifics of how that happens will come down to the importer implementation for a particular simulation format.

Conversely, Ellipsoids and other non-pointlike rigid shapes can't be exported to MDAnalysis.  You could *I suppose* export the center-of-mass positions of each, but those will in general differ from the center of mass position of the atoms underneath it (no coupling besides transformation). 

TL;DR, the representation stores info which neither HOOMD DPD nor MDAnalysis can store faithfully, so it's important to be judicious in what's provided.
We can keep this code-as is without affecting testing, but I'm really trying to make sure you understand what each part of it is doing, since this fixture is lifted verbatim from the examples. An "Ellipsoid" does not refer to the whole repeat unit, but rather an attribute of it.

mupt/interfaces/mdanalysis/exporters.py
timbernat: That sounds reasonable. To be clear then, are you suggesting opting for "molecules", "repeat units", and "particles" within *this* MuPT module as well?

mupt/tests/conftest.py
timbernat: This is fine to stay here for now, but we should revisit it as more tests are added. The build logic will likely be made much more compact near-term, and each set of tests might need to set up different polymer systems, or include different amounts of information in the representation.

README.md
timbernat: @mrshirts has there been any consensus from the PIs on this?

mupt/tests/interfaces/mdanalysis/test_exporters.py
timbernat: > Xfail only checks if the test fails, not that a specific error is thrown

`pytest` test fail _precisely when_ an Exception is raised, often an `AssertionError` raised by an `assert <condition>` clause. `xfail` _does_ allows you to provide a specific type of Exception(s) to check for via the [`raises` keyword](https://docs.pytest.org/en/stable/how-to/skipping.html#raises-parameter), which is what I was suggesting here. Using xfails in parameterized examples allows to you be much more specific about which exceptions and conditions fail tests, as the [docs indicate](https://docs.pytest.org/en/7.1.x/how-to/assert.html#:~:text=Note%20that%20it%20is%20also%20possible%20to%20specify%20a%20%E2%80%9Craises%E2%80%9D%20argument%20to%20pytest.mark.xfail%2C%20which%20checks%20that%20the%20test%20is%20failing%20in%20a%20more%20specific%20way%20than%20just%20having%20any%20exception%20raised%3A).

mupt/interfaces/mdanalysis/exporters.py
```diff
@@ -0,0 +1,427 @@
...
+    # But since atom handles like ('ATOM', 14) are unique WITHIN a residue, we need more info
+    # Strategy: Store both the residue object and its atom handle-to-global-index mapping
+    residue_to_atom_global_idx = []  # List of dicts: residue_to_atom_global_idx[global_res_idx][atom_handle] = global_atom_idx
+
+    # Counters
+    atom_idx = 0
+    res_idx = 0
+
+    # ----------------------------
+    # Traverse hierarchy explicitly DEV: JRL -> Consider implementing a more efficient traversal like DFS in the future
```
timbernat, line 170: As pointed out, we want a more transferrable method of traversal and export of general hierarchies, but this will do for now

mupt/tests/conftest.py
```diff
@@ -0,0 +1,613 @@
...
+    total_atoms = len(univprim.leaves)
+    total_residues = sum(len(chain.children) for chain in univprim.children)
+    logger.info(
+        f"Built system: {n_chains} chains, {total_residues} residues, {total_atoms} atoms"
+    )
+
+    return univprim
+
+
+"""
```
timbernat: Not sure what this free-floating docstring is doing *here*?

mupt/tests/conftest.py
```diff
@@ -0,0 +1,613 @@
...
+            polyethersulfone_smiles,
+            mid_distrib,
+            n_chains=n_chains,
+            chain_len_min=chain_len_min,
+            chain_len_max=chain_len_max,
+            random_seed=random_seed,
+            **kwargs,
+        )
+
+    return _make_PES
```
timbernat: What's the reason for making these higher-order factory functions? The fixtures that end up depending on them return Primitive objects, not the wrapped functions themselves, and could just call `build_polymer_system()` directly. Seems like a lot of extra complexity with no clear benefit

mupt/tests/conftest.py
```diff
@@ -0,0 +1,613 @@
...
+@pytest.fixture
+def helium_resname_map() -> dict[str, str]:
+    """Residue name mapping for Helium SAAMR system."""
+    return {"He_unit": "HEL"}
+
+
+@pytest.fixture
+def single_helium_atom_saamr() -> Primitive:
+    """
+    Fixture providing the simplest possible SAAMR-compliant system: a single Helium atom.
```
timbernat, line 428: The fact that it's Helium is a red herring, as what matter here is just that the system contains a single atom.

README.md
timbernat, line 51: @mrshirts has there been any consensus from the PIs on this?

timbernat (approved):

mupt/interfaces/mdanalysis/exporters.py
joelaforet: No, this module reads data from a MuPT representation (with MuPT terminology) and places the data into containers that are then read by MDAnalysis. I am saying that the containers that will be used by MDAnalysis should have terminology associated with MDAnalysis.

The names for the containers were chosen intentionally to match the constructor for an empty MDAnalysis Universe: [here](https://github.com/MuPT-hub/mupt/pull/24/changes#diff-a81f6072f5982aec36d4353a05d13948643e2f7189f936861fee45d061dbb778R385-R392)

mupt/tests/conftest.py
joelaforet: Concerns are addressed by commit ff7373c

joelaforet (commented):
Comments are addressed, ready for merge.

PR #50, opened 2026-03-30
refactor(mdanalysis): role-based export with Strategy pattern for arbitrary-depth trees

### LINKED ISSUE #19: How to parse a MuPT representation for export to MDAnalysis?
## Background

I am attempting to write an export utility that creates an MDAnalysis universe from a MuPT representation of an all-atom system. The general recipe for creating a MDAnalysis universe requires the following information to be assigned explicitly:

1. Number of atoms (int), how many `atoms` are in the system
2. Number of residues (int), how many `residues` are in the system. In this context, a `residue` is defined as a grouping of `atoms` such that the grouping composes a unit that will be repeated elsewhere in the system. A single CG bead serves the same purpose as a single `residue`. As a concrete example, consider the amino acid Alanine. It is made up of `atoms` that compose a unit that will be repeated throughout a polymer (protein).
3. Number of segments (int), how many distinct, non-covalently bound entities there are in the system. This can also be called `chains`. This is a way of organizing the separate molecules within the system.
4. Atom residue index (list[int]), where length == `number of atoms`. This is a mapping where each atom in the system is assigned an index that corresponds to the residue it is a part of. 
5. Residue segment index (list[int]), where length == `number of residues`. This is a mapping where each `residue` in the system is assigned an index that corresponds to the chain/segment it is a part of.

Once the above information is computed/retrieved, you can instantiate an empty MDAnalysis universe.

## Question 1: How to get bonds out of a primitive that represents an entire system?
After the Universe exists, then you need to add in [Topology Attributes](https://userguide.mdanalysis.org/stable/topology_system.html#topology-attributes) that allow MDAnalysis' selection algebra to work as intended. Importantly, you must now manually specify the bonds between atoms in the system. How does one retrieve information about what atoms are bonded within a MuPT representation? Is the [current implementation correct?](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/mdanalysis/exporters.py#L181-L210)

## Question 2: How does the flattening operation work?
MDAnalysis expects atoms to be stored in a flattened state, where each entry has some information about what segment + residue it belongs to. I.E. all of the atoms in the system will be arranged 1-N. The MuPT representation stores atoms as leaves of a tree. When we flatten this system, how are the atoms now arranged? I iterated over each molecule through [nested for loops](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/mdanalysis/exporters.py#L95C5-L129C31), and expected this to match the ordering of atoms when we flattened a representation. I loaded in the appropriate information into an MDAnalysis universe and added in coordinates generated by flattening the MuPT representation, exporting this to RDKit, and then generating a conformer. The coordinates look correct, however the residue/segment information does not match what I would expect. 

Image

This image is a set of 3 polymer molecules generated from the [`ellipsoidal_chain_placement.ipynb`](https://github.com/MuPT-hub/mupt-examples/blob/main/examples_system/ellipsoidal_chain_placement.ipynb) notebook. I used my current implementation of the [`primitive_to_mdanalysis` ](https://github.com/joelaforet/mupt/blob/MDAexport/mupt/interfaces/mdanalysis/exporters.py#L95C5-L129C31)function to generate a MDAnalysis universe directly from this primitive, added coordinates generated by exporting the primitive to RDKit, and then used the MDAnalysis PDB writer to export the system for visualization. 

You can see in the top bar that the layout of polymer chains is as expected, head->bisp->bisp->tail. This is how one would read a linear polymer from left to right. However, the atoms that are assigned to each residue appear to be spread across different molecules and different residues. Some of the atoms of the dark blue residue are orange.

I would have expected the molecule that starts with the red head to contain all of the colored residues, and then each subunit of the polymer is the same color. We should not have random hydrogens from the orange residue replacing the hydrogens from the dark blue residue.

Can someone take a look at this loop that iterates over the MuPT representation and comment on where any misunderstandings may have occured? 

## How to test this: (minimum reproducing example)
First, acquire the [MDAexport branch of my fork of mupt](https://github.com/joelaforet/mupt/tree/MDAexport). Once that is loaded into your environment, walk through the [`ellipsoidal_chain_placement.ipynb`](url) notebook without changing anything. Before flattening the Primitive generated that contains the whole polymer system, make sure to store it in some variable, `univprim`. Generate coordinates for the polymer system by exporting to RDKit, and the extracting those via something like `coords = univmol.GetConformer(0).GetPositions()`.

Generate a MDAnalysis universe from the coordinates and MuPT representation by:

```
from mupt.interfaces.mdanalysis.exporters import primitive_to_mdanalysis

universe= primitive_to_mdanalysis(univprim, coords=coords)
print(f"Created MDAnalysis Universe with {universe.atoms.n_atoms} atoms")
```

Export this to a PDB file to confirm correct assignment of bonds/residues/segments by:
```
ag = universe.select_atoms('all')

ag.write("MuPT_MDA_Export.pdb")
```

Open this file with some molecular visualizer (like PyMol).

joelaforet:
## Summary
Closes #19.

Replaces the hard-coded 3-level nested-loop MDAnalysis exporter with a role-based, Strategy-pattern architecture that supports Primitive trees of arbitrary depth. The [old exporter](https://github.com/MuPT-hub/mupt/blob/main/mupt/interfaces/mdanalysis/exporters.py#L172-L215) assumed a fixed Universe → Molecule → Repeat-Unit → Atom (SAAMR) hierarchy, which broke for non-standard tree shapes and made extending to coarse-grained or mixed-resolution exports impossible without rewriting the core loop.
## Motivation
Issue #19 surfaced two problems with the original exporter:
1. **Atom-to-residue/segment assignment was coupled to tree depth.** The [nested `for chain / for residue / for atom` loop](https://github.com/MuPT-hub/mupt/blob/main/mupt/interfaces/mdanalysis/exporters.py#L172-L215) only worked for depth-3 trees. A depth-4 tree (e.g., Universe → Domain → Chain → Residue → Atom) could not be exported without restructuring the Primitive tree itself.
2. **`flatten()` produces nondeterministic atom ordering.** [`flatten()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L1015-L1019) repeatedly calls [`expand()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L926-L997) on children returned by [`expandable_children`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L918-L924), which returns a **`set[PrimitiveHandle]`**. Since [Python set iteration order depends on hash randomization](https://docs.python.org/3/reference/datamodel.html#object.__hash__) (`PYTHONHASHSEED`), the order in which children are expanded varies between runs. Each `expand()` call destructively mutates the parent's child registry — it detaches the target child, promotes its grandchildren, and remaps connector handles using the [`UniqueRegistry`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mutils/containers.py) ticker. This means expanding child A before child B produces different handle assignments (and therefore a different [`root.leaves`](https://anytree.readthedocs.io/en/stable/api/anytree.node.html#anytree.node.nodemixin.NodeMixin.leaves) ordering) than expanding B before A. When the old exporter used `flatten()` for atom positions but a deterministic nested loop for residue/segment indices, the two orderings disagreed — atom positions were assigned to incorrect residues, producing the color scrambling shown in #19.

This PR addresses problem (1) directly with role-based export. Problem (2) is a known issue with [`flatten()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L1015-L1019) that is documented but intentionally not fixed here, since the new exporter never calls `flatten()`.
### [`PrimitiveRole`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/roles.py#L16-L40) Enum
Four canonical roles that map to the [MDAnalysis topology hierarchy](https://userguide.mdanalysis.org/stable/topology_system.html):
| Role | MDAnalysis concept | Description |
|------|-------------------|-------------|
| `UNIVERSE` | Universe | Root container |
| `SEGMENT` | Segment | Non-covalently-bonded entity (chain/molecule) |
| `RESIDUE` | Residue | Repeating sub-unit (monomer, CG bead group) |
| `PARTICLE` | Atom | Exportable particle (atom or CG bead) |
Primitives at intermediate depths (between canonical roles) receive no role — they are transparent grouping nodes that the exporter traverses through via [`PreOrderIter`](https://anytree.readthedocs.io/en/stable/api/anytree.iterators.html#anytree.iterators.preorderiter.PreOrderIter).
### Role Assignment ([`properties.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/properties.py))
- [`is_SAAMR_compliant(prim)`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/properties.py#L13-L27) — checks that all leaves are atoms at depth 3.
- [`assign_SAAMR_roles(prim)`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/properties.py#L30-L56) — assigns roles to a SAAMR-compliant hierarchy. For non-SAAMR trees (depth-4+), users assign roles manually.
The exporter **never infers roles from tree depth**. If roles are missing, it raises a `ValueError` pointing users to `assign_SAAMR_roles()` or manual assignment.
### Strategy Pattern ([`strategies.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py))
- [`MDAExportStrategy`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L161-L176) (ABC) — defines the interface for export strategies.
- [`AllAtomExportStrategy`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L178-L352) — the production implementation. Walks the role-tagged tree using [`PreOrderIter`](https://anytree.readthedocs.io/en/stable/api/anytree.iterators.html#anytree.iterators.preorderiter.PreOrderIter) to [discover SEGMENT and RESIDUE nodes at any depth](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L280-L287), maps PARTICLE leaves to global atom indices, and resolves bonds recursively through [`external_connectors`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L432).
- [`CoarseGrainedExportStrategy`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L354-L368) — stub for future CG export.
### Refactored Entry Point ([`exporters.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/exporters.py))
[`primitive_to_mdanalysis()`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/exporters.py#L113) validates SAAMR compliance, auto-assigns roles if missing, and delegates to the strategy. Users can pass an explicit `strategy=` parameter to override the default `AllAtomExportStrategy`.
### Bond Resolution
[`_resolve_to_atom()`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L55-L101) recursively follows `external_connectors` from any tree level down to leaf atoms, with:
- Depth guard (max 50 hops) to [catch connector cycles](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L71-L77)
- `KeyError`/`AttributeError` catches with [actionable error messages](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L79-L84)
- Deterministic output: [`frozenset` pairs are sorted](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L319-L325) by `(primitive_handle, connector_handle)` before unpacking; [final bond list is sorted](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L342-L349) by atom-index pair
## Changes
**New files (2):**
- [`mupt/mupr/roles.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/roles.py) — `PrimitiveRole` enum
- [`mupt/interfaces/mdanalysis/strategies.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py) — Strategy ABC + implementations
**Modified files (7):**
- [`mupt/mupr/primitives.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/primitives.py#L114-L201) — Added [`role`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/primitives.py#L173-L181) field, [`is_universe`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/primitives.py#L199-L201) property, [`__copy__`/`__deepcopy__`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/primitives.py#L1094) role propagation
- [`mupt/mupr/properties.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/properties.py) — Added [`is_SAAMR_compliant()`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/properties.py#L13-L27) and [`assign_SAAMR_roles()`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/properties.py#L30-L56)
- [`mupt/mupr/__init__.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/__init__.py) — Exported `PrimitiveRole`
- [`mupt/interfaces/mdanalysis/exporters.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/exporters.py) — Refactored to delegate to strategy
- [`mupt/interfaces/mdanalysis/__init__.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/__init__.py) — Public API: `primitive_to_mdanalysis`, `MDAExportStrategy`, `AllAtomExportStrategy`
- [`mupt/tests/conftest.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/tests/conftest.py) — Fixtures auto-tag SAAMR roles; added [`depth4_helium_system`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/tests/conftest.py#L467) and [`depth4_bonded_system`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/tests/conftest.py#L532)
- [`mupt/tests/interfaces/mdanalysis/test_exporters.py`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/tests/interfaces/mdanalysis/test_exporters.py) — 44 tests
**Net**: 9 files changed, +1185, −363
## Test Coverage
44 MDAnalysis-specific tests across 5 categories:
| Category | Tests | What they verify |
|----------|-------|-----------------|
| SAAMR export parity | 14 | Atom count, bond count, segment count across PE 2-mer, 3-mer, 10-chain, PES, helium |
| Role enforcement | 7 | Rejection of unroled trees, missing SEGMENT role, unroled leaves, nested SEGMENT/RESIDUE, empty trees, non-SAAMR trees |
| Strategy API | 3 | Explicit `AllAtomExportStrategy` produces identical results to default path |
| Depth-4 unbonded | 6 | Atom/segment/residue counts, resnames, element names, zero bonds |
| Depth-4 bonded | 10 | Bond count, bond-order values, primitive-level bond count parity, inter-residue bond spans residues, inter-residue bond connects correct atom types, residue atom distribution |
Full suite: **191 passed, 1 skipped, 3 xfailed, 0 failed**.
## Performance
The new exporter is ~2x slower than the old one (constant factor, both O(N)). This is an intentional trade-off for validation, safety, and extensibility:
| System | Atoms | Old (ms) | New (ms) | Ratio |
|--------|------:|----------:|----------:|------:|
| PE small | 140 | 1.6 | 3.1 | 1.9× |
| PE medium | 2,800 | 25.2 | 51.6 | 2.0× |
| PE large | 11,600 | 104.8 | 214.8 | 2.1× |
| PES small | 1,119 | 9.6 | 19.6 | 2.0× |
| PES medium | 7,092 | 58.7 | 128.0 | 2.2× |
| PES large | 21,042 | 177.6 | 377.9 | 2.1× |
The overhead comes from [`is_SAAMR_compliant`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/properties.py#L13-L27) validation, [`assign_SAAMR_roles`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/mupr/properties.py#L30-L56), strategy [`validate()`](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L186-L249) with 3 [`PreOrderIter`](https://anytree.readthedocs.io/en/stable/api/anytree.iterators.html#anytree.iterators.preorderiter.PreOrderIter) passes, and the role-filtered [hierarchy walk](https://github.com/joelaforet/mupt/blob/c714aa5/mupt/interfaces/mdanalysis/strategies.py#L280-L310). For a 21K-atom system the total export time is ~378ms.
## Known Issues
- **[`flatten()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L1015-L1019) nondeterminism**: [`expandable_children`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L918-L924) returns a `set[PrimitiveHandle]`, and [`flatten()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L1015-L1019) iterates over this set to decide which children to [`expand()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L926-L997). Since [Python randomizes hash seeds per-process](https://docs.python.org/3/reference/datamodel.html#object.__hash__), set iteration order is unpredictable, and each `expand()` destructively mutates handle assignments — so the final leaf ordering depends on expansion order. This affects any consumer that relies on `flatten()` ordering (including the RDKit exporter), but does **not** affect the MDAnalysis exporter introduced in this PR, which never calls `flatten()`. Not addressed in this PR.


mupt/interfaces/mdanalysis/strategies.py
```diff
@@ -0,0 +1,368 @@
...
+    _max_depth: int = 50,
+) -> Primitive:
+    """Recursively follow external_connectors to find the leaf atom.
+
+    Raises
+    ------
+    ValueError
+        If the connector chain is broken (missing child or external
+        connector entry) or if recursion exceeds *_max_depth*, which
+        indicates a cycle in the connector graph.
```
timbernat: (clarification) Can you explain what you mean by this? The Connectors themselves don't form a graph, but moreover the kind of cycle you refer to here seems to mean one of the child nodes is also its own ancestor which doesn't make sense.

mupt/interfaces/mdanalysis/strategies.py
```diff
@@ -0,0 +1,368 @@
...
+    connector = child.connectors.get(conn_ref.connector_handle)
+    if connector is None:
+        LOGGER.warning(
+            "Connector %s not found on child %s of parent %s; "
+            "defaulting bond order to 1.0",
+            conn_ref.connector_handle,
+            conn_ref.primitive_handle,
+            parent.label,
+        )
+        return 1.0
```
timbernat: (replacement) This whole block could be a one-line call, via
```suggestion
connector = parent.fetch_connector_on_child(conn_ref)
```
which will raise a descriptive `MissingConnectorError` if none is found

mupt/interfaces/mdanalysis/strategies.py
```diff
@@ -0,0 +1,368 @@
...
+
+    bondtype: Optional[str] = getattr(connector, "bondtype", None)
+    order = BOND_ORDER.get(bondtype, 1.0)
```
timbernat: (clarification) Why default to single bonds when Connectors are not present, rather than say 0-order bonds?

mupt/interfaces/mdanalysis/strategies.py
```diff
@@ -0,0 +1,368 @@
...
+    residue_names: list[str] = field(default_factory=list)
+    residue_segindex: list[int] = field(default_factory=list)
+    residue_ids: list[int] = field(default_factory=list)
+    bonds: list[tuple[int, int]] = field(default_factory=list)
+    bond_orders: list[float] = field(default_factory=list)
+    bonds_set: set[tuple[int, int]] = field(default_factory=set)
+    num_segments: int = 0
+
+
+class MDAExportStrategy(ABC):
```
timbernat, line 148: (compliment) ABCs are a good choice here!

mupt/mupr/roles.py
```diff
@@ -0,0 +1,40 @@
+"""
+Canonical roles that Primitives can play in an exportable hierarchy.
```
timbernat: (naming) "Canonical" is only the right term for SAAMR; there may be many other designations of Primitives (e.g. "functional group", "Kuhn unit", "sticker/spacer") which we don't want to need to keep updating.

mupt/roles.py
```diff
@@ -0,0 +1,40 @@
+"""
+Canonical roles that Primitives can play in an exportable hierarchy.
+
+These roles allow Primitives to be explicitly tagged with their semantic
+purpose within a molecular representation, enabling generic tree traversal
+for export to external toolkits (e.g., MDAnalysis, RDKit) without
+hard-coding assumptions about tree depth or structure.
```
timbernat, line 7: (suggestion) It's not obvious at-a-glance that this is really the case; from the `mdanalysis.strategies' `source it becomes clear that any Primitive with no or duplicate role get skipped over, but the 4 categories here appear as though enforcing a 4-level SAAMR. An explicit `UNASSIGNED` role would make this option clearer, as well as avoiding the `NoneType` sentinel check in all downstream code

mupt/mupr/primitives.py
```diff
@@ -110,6 +111,7 @@ def __init__(
         children : Optional[Iterable['Primitive']]=None,
         label : Optional[PrimitiveLabel]=None,
         metadata : Optional[dict[Hashable, Any]]=None,
+        role : Optional[PrimitiveRole]=None,
```
timbernat: (remove) These roles are pretty useful for SAAMR/MDAnalysis export, but we don't want to be hard-coding them directly into the core representation, as that sort of defeats the scale-agnostic internals of the representation. There are other ways that could be made stronger (e.g. I'm planning on deprecating `element` on anything but atomic Primitives), but this shouldn't be directly in Primitive source. Assigning roles in Primitive metadata would be preferred, especially since they're optional.

mupt/interfaces/mdanalysis/strategies.py
```diff
@@ -0,0 +1,368 @@
...
+                    "AllAtomExportStrategy requires atomic PARTICLE leaves."
+                )
+            atom_symbol = atom.element.symbol
+            data.atom_elements.append(atom_symbol)
+            data.atom_names.append(atom_symbol)
+
+            if hasattr(atom, "shape") and atom.shape is not None:
+                data.atom_positions.append(list(atom.shape.centroid))
+            else:
+                data.atom_positions.append([0.0, 0.0, 0.0])
```
timbernat: (amend) Would be good to have the default atom position be passed as an optional argument (as [in the RDKit exporter](https://github.com/MuPT-hub/mupt/blob/main/mupt/interfaces/rdkit/exporters.py#L42)), or possibly a CAllable that acts on the atom.

mupt/mupr/properties.py
```diff
@@ -15,6 +16,41 @@ def is_SAAMR_compliant(prim: Primitive) -> bool:
...
+   hierarchy requires all leaves to be atoms at depth 3.  Note that
+   ``prim.leaves`` always includes at least ``prim`` itself (anytree
+   treats childless nodes as their own leaf), so the ``all()`` predicate
+   is never evaluated over an empty iterable.
    """
 
    return all(leaf.is_atom and (leaf.depth == 3) for leaf in prim.leaves)
+
+
+def assign_SAAMR_roles(prim : Primitive) -> None:
```
timbernat: (move) This may be more at-home in `roles.py` or even `mdanalysis.strategies.py`, since it's assigning many sub-primitives properties, rather than determining one from the whole tree.

mupt/tests/interfaces/mdanalysis/test_exporters.py
```diff
@@ -232,3 +256,396 @@ def test_invalid_resname_map_raises_value_error(primitive_fixture, resname_map,
...
+    # Bond comparison — handle the no-bonds case (e.g. single helium atom)
+    default_has_bonds = hasattr(mda_default, "bonds") and hasattr(mda_default.atoms, "_topology") and "bonds" in mda_default.atoms._topology.attrs
+    explicit_has_bonds = hasattr(mda_explicit, "bonds") and hasattr(mda_explicit.atoms, "_topology") and "bonds" in mda_explicit.atoms._topology.attrs
+    if default_has_bonds and explicit_has_bonds:
+        assert len(mda_default.bonds) == len(mda_explicit.bonds)
+    else:
+        assert default_has_bonds == explicit_has_bonds
+
+
+def test_strategy_rejects_unroled_primitives():
```
timbernat: (naming, optional) "Unroled"? Perhaps `role_free`

mupt/interfaces/mdanalysis/exporters.py
```diff
@@ -18,410 +22,182 @@
...
-            LOGGER.info(f"Added {len(bonds)} bonds (no bond orders)")
-    else:
-        LOGGER.info("No bonds to add")
+    >>> from mupt.interfaces.mdanalysis.strategies import AllAtomExportStrategy
+    >>> strategy = AllAtomExportStrategy()
+    >>> universe = primitive_to_mdanalysis(my_univprim, resname_map, strategy=strategy)
+    """
+    if strategy is None:
+        # Backward-compatible path: require SAAMR, auto-assign roles
+        if not is_SAAMR_compliant(univprim):
```
timbernat: (comment) Similar comment as in `properties.py`, SAAMR is still forced, not arbitrary tree as claimed

mupt/interfaces/mdanalysis/exporters.py
timbernat, line 1: (compliment) Much cleaner!

mupt/interfaces/mdanalysis/strategies.py
```diff
@@ -0,0 +1,368 @@
...
+        LOGGER.warning(
+            "Connector %s not found on child %s of parent %s; "
+            "defaulting bond order to 1.0",
+            conn_ref.connector_handle,
+            conn_ref.primitive_handle,
+            parent.label,
+        )
+        return 1.0
+
+    bondtype: Optional[str] = getattr(connector, "bondtype", None)
```
timbernat: (replacement) Connector.bondtype is one of the few non-optional fields in that class and defaults to BondType.UNSPECIFIED, so the NoneType check shouldn't be necessary. A much simpler implementation would look like:
```suggestion
return BOND_ORDER[connector.bondtype)
```
which will raise KeyError if bondtype is either None or not among the pre-compiled list of types

mupt/mupr/properties.py
```diff
@@ -15,6 +16,41 @@ def is_SAAMR_compliant(prim: Primitive) -> bool:
...
+   prim.role = PrimitiveRole.UNIVERSE
+
+   for segment in prim.children:
+      segment.role = PrimitiveRole.SEGMENT
+
+      for residue in segment.children:
+         residue.role = PrimitiveRole.RESIDUE
+
+         for particle in residue.children:
+            particle.role = PrimitiveRole.PARTICLE
```
timbernat: (comment) This just moves the assumption of a laminar 4-level tree into this function, rather than supporting arbitrary depth as claimed. Not necessarily a bad thing, but pointing out that the changes don't work out-of-box on any hierarchies other than SAAMRs.

timbernat (changes requested):
Overall looks pretty good, much cleaner than the prior draft! Export in this revision is still bound to SAAMR (not arbitrary-depth as advertised), but that's is more-than-sufficient for MDAnalysis-related functionality. A few cuts, moves, and tweaks recommended, and then ready to merge once those are made

mupt/mupr/primitives.py
joelaforet: Changed to set roles as part of the primitive's metadata [here](https://github.com/MuPT-hub/mupt/pull/50/changes#diff-a5695cc7bd9ec4b03a637a8b683c098d922c63850c700838c4be53805b194789R145-R146)

mupt/mupr/roles.py
joelaforet: I used "Canonical" in anticipation of this file housing other roles as we build more exporters. I agree that this file only contains SAAMR roles, so "Canonical" may be a misnomer, however this file will be extended in the future.

mupt/roles.py
joelaforet: Added an `UNASSIGNED` role [here](https://github.com/MuPT-hub/mupt/pull/50/changes#diff-59c9b6cc78e9a543da549112a5d08bea37b92e23407bb70901c53375c144efd5R43) and as a default on Primitive construction [here](https://github.com/MuPT-hub/mupt/pull/50/changes#diff-a5695cc7bd9ec4b03a637a8b683c098d922c63850c700838c4be53805b194789R174)

timbernat (approved):
Looks ready to go

PR #55, opened 2026-05-01
Improved BoundedShapes and array typing

### LINKED ISSUE #10: Feature additions to BoundedShape and subtypes
Currently, [BoundedShape](https://github.com/MuPT-hub/mupt/blob/main/mupt/geometry/shapes.py#L77) supports [PointCloud](https://github.com/MuPT-hub/mupt/blob/main/mupt/geometry/shapes.py#L109), [Sphere](https://github.com/MuPT-hub/mupt/blob/main/mupt/geometry/shapes.py#L158), and [Ellipsoid](https://github.com/MuPT-hub/mupt/blob/main/mupt/geometry/shapes.py#L216) as concrete child types which implement the contracts BoundedShape lays out (viz. centroid(), volume(), and contains()). 

Have plans to enhance the capability

### New BoundedShape contracts
* support() - which calculates the furthest point in a given direction on the surface of a body,
  * bounding_box() - tight XYZ bounding box; trivial to implement if support() is implemented
* surface_mesh() - for providing consistent interface when plotting Shapes; Sphere and Ellipsoid actually already implement this, and the only task here would be to implement an analogue for the convex_hull information associated with PointCloud instances

### New BoundedShape subtypes
* Cylinder/Rod - does what it says on the can
* NullShape - a subtype of which instances _can_ be instantiated (unlike the ABC BoundedShape base), but which raises Exception on all contracts; used as alternative to Optional[BoundedShape] to simplify type checks _DEV: not clear whether this is necessary?_
* Others worth including?

### Additional functionality for existing Shape types
* PointCloud - provide support for arbitrary labels to index individual positions by (rather than just integers) _DEV: ought to weight speed vs flexibility on this one_

timbernat:
## Description
This PR aims to improve consistency, usability, and feature completeness of the `mupt.geometry` module, especially as pertains to `BoundedShape`, as well as strengthening array-related typing.

Associated with the related [`better-shapes` branch of `mupt-examples`](https://github.com/MuPT-hub/mupt-examples/tree/better-shapes), which greatly simplifies and enhances the [mupt-shapes.ipynb example](https://github.com/MuPT-hub/mupt-examples/blob/better-shapes/examples_repr/mupt_shapes.ipynb)

## Todos
  - [x] Resolve #10
  - [x] Standardize abstract interface for Shapes
    - [x] Provide uniform contract for surface_mesh() (in terms of a triangulated mesh)
    - [x] ~Implement support vector calculation~ Split off into #63 in the interest of time
    - [x] Implement comparisons between instances of each type of shape
    - [x] Implement uniform scaling interface
    - [x] Add `visualize()` interface with thin mpl wrapper
    - [x] Add Cylinder/Rod class
  - [x] Decouple BoundedShape from RigidlyTransformable
  - [x] Tighten up numpy-related typehints
  - [x] Fix non-updating visualization for PointCloud in `mupt-examples` 
  - [x] Write tests
    - [x] Containment 
    - [x] Volume
    - [x] Equality
    - [x] Scaling 
    - [x] Rigid transformations

## Questions
- [x] ~Should a NullShape class be implemented to avoid NoneType sentinels for unset `.shape` fields on Primitive?~
- [x] ~Is it worth adding support for arbitrary mapped labels to points in `PointCloud.positions`, rather than just int indices?~
_DEV: opting for no on both questions_

## Status
- [x] Address points in #10 
- [x] Update [mupt_shapes notebook](https://github.com/MuPT-hub/mupt-examples/blob/main/examples_repr/mupt_shapes.ipynb) in parallel examples repo
- [x] Ensure new tests pass
- [x] Pass review
- [x] Ready to go
  - [x] Close [examples #24](https://github.com/MuPT-hub/mupt-examples/pull/24) once merged

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Run example notebook tests
         run: |
-          git clone -b better-shapes https://github.com/MuPT-hub/mupt-examples 
+          git clone https://github.com/MuPT-hub/mupt-examples 
           pushd mupt-examples
           source notebook_tests_master.sh
           popd
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Run example notebook tests
         run: |
-          git clone https://github.com/MuPT-hub/mupt-examples
+          git clone -b better-shapes https://github.com/MuPT-hub/mupt-examples 
           pushd mupt-examples
           source notebook_tests_master.sh
           popd
--- a/mupt/geometry/arraytypes.py
+++ b/mupt/geometry/arraytypes.py
@@ -4,57 +4,60 @@ __author__ = 'Timotej Bernat'
 __email__ = 'timotej.bernat@colorado.edu'
 
 from typing import (
-    Annotated,
-    Generic,
     Literal,
     Optional,
     TypeVar,
+    Union,
 )
 S = TypeVar('S') # pure generics
 T = TypeVar('T') # pure generics
 
 import numpy as np
 import numpy.typing as npt
-from numpy import ndarray
 from numbers import Number, Real
 
 
 # Numeric typehints
-Numeric = TypeVar('Numeric', bound=Number) # typehint a number-like generic type
-RealValued = TypeVar('RealValued', bound=Real)
+NumberLike = Union[np.number, Number, float] # DEV: stupidly, but "float" does not typehint as Number in static type checkers, so have to add it manually
+Numeric = TypeVar('Numeric', bound=Number)
+NumericNP = TypeVar('NumericNP', bound=np.dtype[np.number])
+
+# RealValued = TypeVar('RealValued', bound=Real)
+# RealValuedNP = TypeVar('RealValuedNP', bound=np.dtype[np.floating])
 
 # Numpy array type annotations
-Shape = tuple # the shape field of a numpy array
-DType = TypeVar('DType', bound=np.generic) # the data type of a numpy array
+Shape = tuple
+DType = TypeVar('DType', bound=np.dtype)
 
+## Typehints for indeterminate size of a given array dimension
+M = TypeVar('M', bound=int) 
+N = TypeVar('N', bound=int)
+P = TypeVar('P', bound=int)
 Dims = TypeVar('Dims', bound=int) # intended to typehint the number of dimensions
 DimsPlus = TypeVar('DimsPlus', bound=int) # intended to typehint the number of dimensions +1 (no easy way to do arithmetic to generic types yet)
-M = TypeVar('M', bound=int) # typehint the size of a given dimension
-N = TypeVar('N', bound=int) # typehint the size of a given dimension
-P = TypeVar('P', bound=int) # typehint the size of a given dimension
 
 # Fixed-size vector and array type annotations - consider deprecating, since they're not currently being used anywhere
 ## DEV: this type of hard-coding sucks, but is the best we can do with the current Python type system
-Vector2  = np.ndarray[Shape[Literal[2]], Numeric]
-Vector3  = np.ndarray[Shape[Literal[3]], Numeric]
-Vector4  = np.ndarray[Shape[Literal[4]], Numeric]
-VectorN  = np.ndarray[Shape[N], Numeric]
+Vector2  = np.ndarray[Shape[Literal[2]], NumericNP]
+Vector3  = np.ndarray[Shape[Literal[3]], NumericNP]
+Vector4  = np.ndarray[Shape[Literal[4]], NumericNP]
+VectorN  = np.ndarray[Shape[N], NumericNP]
 
-Array2x2 = np.ndarray[Shape[Literal[2], Literal[2]], Numeric]
-Array3x3 = np.ndarray[Shape[Literal[3], Literal[3]], Numeric]
-Array4x4 = np.ndarray[Shape[Literal[4], Literal[4]], Numeric]
+Array2x2 = np.ndarray[Shape[Literal[2], Literal[2]], NumericNP]
+Array3x3 = np.ndarray[Shape[Literal[3], Literal[3]], NumericNP]
+Array4x4 = np.ndarray[Shape[Literal[4], Literal[4]], NumericNP]
 
-ArrayNx2 = np.ndarray[Shape[N, Literal[2]], Numeric]
-ArrayNx3 = np.ndarray[Shape[N, Literal[3]], Numeric]
-ArrayNx4 = np.ndarray[Shape[N, Literal[4]], Numeric]
+ArrayNx2 = np.ndarray[Shape[N, Literal[2]], NumericNP]
+ArrayNx3 = np.ndarray[Shape[N, Literal[3]], NumericNP]
+ArrayNx4 = np.ndarray[Shape[N, Literal[4]], NumericNP]
 
-Array2xN = np.ndarray[Shape[Literal[2], N], Numeric]
-Array3xN = np.ndarray[Shape[Literal[3], N], Numeric]
-Array4xN = np.ndarray[Shape[Literal[4], N], Numeric]
+Array2xN = np.ndarray[Shape[Literal[2], N], NumericNP]
+Array3xN = np.ndarray[Shape[Literal[3], N], NumericNP]
+Array4xN = np.ndarray[Shape[Literal[4], N], NumericNP]
 
-ArrayNxN = np.ndarray[Shape[N, N], Numeric]
-ArrayNxM = np.ndarray[Shape[N, M], Numeric]
-ArrayMxN = np.ndarray[Shape[M, N], Numeric]
+ArrayNxN = np.ndarray[Shape[N, N], NumericNP]
+ArrayNxM = np.ndarray[Shape[N, M], NumericNP]
+ArrayMxN = np.ndarray[Shape[M, N], NumericNP]
 
 
 # vector comparison
@@ -68,8 +71,8 @@ def as_n_vector(vectorlike : np.ndarray[Shape[N], DType], n : N=3) -> np.ndarray
     return vectorlike.reshape(n)
 
 def compare_optional_positions(
-    position_1 : Optional[np.ndarray[Shape[N], float]],
-    position_2 : Optional[np.ndarray[Shape[N], float]],
+    position_1 : Optional[VectorN],
+    position_2 : Optional[VectorN],
     **kwargs,
 ) -> bool:
     '''Check that two positional values are either 1) both undefined, or 2) both defined and equal'''
--- a/mupt/geometry/arraytypes.py
+++ b/mupt/geometry/arraytypes.py
@@ -66,14 +66,30 @@ TriangulationIndices = np.ndarray[Shape[N, Literal[3]], np.dtype[np.integer]]
 BitVectorN = np.ndarray[Shape[N], BoolNP]
 
 # vector comparison
-def as_n_vector(vectorlike : np.ndarray[Shape[N], DType], n : N=3) -> np.ndarray[Shape[N], DType]:
-    '''Interpret array as a 1D n-element vector''' 
+def as_n_vector(
+    vectorlike : VectorN | Array1xN | ArrayNx1,
+    dimension : int=3,
+    dtype : Optional[npt.DTypeLike]=None,
+) -> VectorN:
+    '''
+    Convert row vector, column vector, Nx1 array, or 1xN array into
+    1D column vector with appropriate dimension and data type
+    
+    Enables permissive ingestion of vector-shaped objects
+    '''
     if not isinstance(vectorlike, np.ndarray): # TODO: include support for list/tuple-like WITHOUT including sets, str, etc
         raise TypeError(f'Vectorlike must be a numpy array, not {type(vectorlike)}')
-    if len(vectorlike) != n:
-        raise ValueError(f'Expected {n}-element vectorlike, received {len(vectorlike)}-element array instead')
     
-    return vectorlike.reshape(n)
+    vector_column = np.atleast_2d(vectorlike).reshape(-1) # permits transposed and nested vector inputs
+    if vector_column.shape != (dimension,):
+        raise ValueError(
+            f'Expected vector with shape {(dimension,)}, got {vector_column.shape}'
+        )
+    
+    if dtype is not None:
+        vector_column = vector_column.astype(dtype)
+        
+    return vector_column
 
 def compare_optional_positions(
     position_1 : Optional[VectorN],
--- a/mupt/geometry/arraytypes.py
+++ b/mupt/geometry/arraytypes.py
@@ -21,7 +21,7 @@ from numbers import Number, Real
 NumberLike = Union[np.number, Number, float] # DEV: stupidly, but "float" does not typehint as Number in static type checkers, so have to add it manually
 Numeric = TypeVar('Numeric', bound=Number)
 NumericNP = TypeVar('NumericNP', bound=np.dtype[np.number])
-BoolNP = TypeVar('BoolNP', bound=np.dtype[np.bool])
+BoolNP = TypeVar('BoolNP', bound=np.dtype[np.bool_])
 
 # RealValued = TypeVar('RealValued', bound=Real)
 # RealValuedNP = TypeVar('RealValuedNP', bound=np.dtype[np.floating])
--- a/mupt/geometry/arraytypes.py
+++ b/mupt/geometry/arraytypes.py
@@ -9,12 +9,13 @@ from typing import (
     TypeVar,
     Union,
 )
-S = TypeVar('S') # pure generics
-T = TypeVar('T') # pure generics
+# pure generics
+S = TypeVar('S')
+T = TypeVar('T')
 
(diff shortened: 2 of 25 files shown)


.github/workflows/CI.yaml
```diff
@@ -66,7 +66,7 @@ jobs:
 
       - name: Run example notebook tests
         run: |
-          git clone https://github.com/MuPT-hub/mupt-examples
+          git clone -b better-shapes https://github.com/MuPT-hub/mupt-examples
```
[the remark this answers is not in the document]
timbernat: I'm aware of this and am waiting for final approval before changing back, as pointing to the main branch of `mupt-examples` will cause CI to fail to to related changes in the `better-shapes` branch of that repo (specifically to [examples_rep/mupt_shapes.ipynb](https://github.com/MuPT-hub/mupt-examples/blob/better-shapes/examples_repr/mupt_shapes.ipynb))

mupt/geometry/shapes/ellipsoid.py
```diff
@@ -0,0 +1,321 @@
...
+    @property
+    def basis(self) -> Array4x4:
+        '''The basis matrix of the Ellipsoid - alias for Ellipsoid.as_affine_matrix()'''
+        return self.as_affine_matrix()
+    
+    @property
+    def principal_axes(self) -> Array3x3:
+        '''The principal axes of the ellipsoid, represented as a 3x3 matrix
+        whose rows are the axis vectors emanating from the Ellipsoid's center'''
+        return self.cumulative_transformation.apply(self.scaling_matrix(as_affine=False))
```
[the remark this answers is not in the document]
timbernat: This is incorrect, as both the translation and rotation component are needed to place the axis-aligned principal axes into place following the accumulated transformation. The use of apply() here is precisely to implement that transformation

mupt/geometry/shapes/shapes.py
```diff
@@ -0,0 +1,123 @@
...
+    def congruent_to(self, other : Self) -> bool:
+        '''Check if another BoundedShape instance has the same size and shape as this one'''
+        ...
+
+    # NB: deliberately NOT abstract - supplies implementation in case of explicit inheritance
+    def __eq__(self, other : Self) -> bool:
+        # DEV: wrapped here to have concrete subclass impls invoked by super().__eq__
+        if not isinstance(other, BoundedShape):
+            raise TypeError(f'Cannot compare BoundedShape to object of type {type(other).__name__}')
+        return self.congruent_to(other)
```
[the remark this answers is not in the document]
timbernat: Fixed in commit [2c5fda8](https://github.com/MuPT-hub/mupt/pull/55/commits/2c5fda8603e61897b16fe9a999fce7e86c82e0c8)

mupt/geometry/shapes/ellipsoid.py
```diff
@@ -0,0 +1,321 @@
...
+    def volume(self) -> float:
+        return 4/3 * np.pi * self.radius**3
+    
+    def contains(self, points : Vector3 | ArrayNxN) -> BitVectorN:
+        return (
+            np.linalg.norm(
+                np.atleast_2d(points - self.center),
+                axis=1, # TODO: 
+            ) <= self.radius
+        ).astype(object)
```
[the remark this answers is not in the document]
timbernat: Fixed in commit [2c5fda8](https://github.com/MuPT-hub/mupt/pull/55/commits/2c5fda8603e61897b16fe9a999fce7e86c82e0c8)

mupt/geometry/shapes/cylinder.py
```diff
@@ -0,0 +1,295 @@
...
+        return normalized(self.axis)
+    
+    @property
+    def face_centers(self) -> np.ndarray[
+        Shape[Literal[3], Literal[2]],
+        np.dtype[np.floating],
+    ]:
+        '''The absolute positions of the midpoints of the leading and tailing faces on the cylinder'''
+        return np.vstack((self.face_center_top, self.face_center_bottom))
+
```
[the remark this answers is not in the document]
timbernat: Fixed in commit [2c5fda8](https://github.com/MuPT-hub/mupt/pull/55/commits/2c5fda8603e61897b16fe9a999fce7e86c82e0c8)

mupt/geometry/shapes/ellipsoid.py
```diff
@@ -0,0 +1,321 @@
...
+    def contains(self, points : Vector3 | ArrayNxN) -> BitVectorN:
+        # Reduce containment check to comparison with auxiliary unit sphere
+        # NB: not applying self.inverse to points because the Ellipsoid basis
+        # matrix in general not a rigid transformation because of axial stretching
+        return ( 
+            np.linalg.norm( 
+                np.atleast_2d(self.resetting_transformation.apply(points) / self.radii), 
+                axis=1,
+            ) <= 1
+        ).astype(object) # need to cast from numpy bool to Python bool
```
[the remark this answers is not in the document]
timbernat: Fixed in commit [2c5fda8](https://github.com/MuPT-hub/mupt/pull/55/commits/2c5fda8603e61897b16fe9a999fce7e86c82e0c8)

mupt/mupr/primitives.py
```diff
@@ -1061,7 +1061,7 @@ def shape(self, new_shape : BoundedShape) -> None:
     def _copy_untransformed(self) -> 'Primitive':
         '''Return a new Primitive with the same information and children as this one, but which has no parent'''
         clone_primitive = self.__class__(
-            shape=(None if self.shape is None else self.shape.copy()),
+            shape=(None if self.shape is None else self.shape.copy()), # DEV TODO: should be _copy_untransformed()?
```
dwhswenson: Not sure that making the copy here is necessary. When you call `__init__`, it sets `shape` via the property, which makes a copy.

mupt/geometry/transforms/rigid/__init__.py
```diff
@@ -4,6 +4,38 @@
...
+    """
+    Generate a random rigid transformation, with both translation and rotation components by default
+
+    Parameters
+    ----------
+    centered : bool, default False
+        Whether the transformation should contain a translational component
+    translation_bound : float, default 1.0
+        Uniform bound along all Cartesian axes for translation component
+        E.g. translation_bound=5.0 will pick a random translation vector from the set [-5, 5]**3
```
dwhswenson: What's the reason for having `centered` as a separate attribute from `translation_bound`? Why not just have `translation_bound=0.0` mean centered (and probably even be the default, since the nonzero default is arbitrary)?

Otherwise, you need code to cover the case that someone sets `centered=True` and also sets a value for `translation_bound`.

mupt/geometry/arraytypes.py
```diff
@@ -4,72 +4,96 @@
...
+ArrayMxN = np.ndarray[Shape[M, N], NumericNP]
 
+TriangulationIndices = np.ndarray[Shape[N, Literal[3]], np.dtype[np.integer]]
+BitVectorN = np.ndarray[Shape[N], BoolNP]
 
 # vector comparison
-def as_n_vector(vectorlike : np.ndarray[Shape[N], DType], n : N=3) -> np.ndarray[Shape[N], DType]:
-    '''Interpret array as a 1D n-element vector''' 
+def as_n_vector(
+    vectorlike : VectorN | Array1xN | ArrayNx1,
```
dwhswenson: Code isn't actually as strict as the type hint here. It'll work with anything as long as the total number of elements is `dimension`. So you could have shape `(1, 3, 1)` with `dimension=3`, or you could have shape `(2, 2)` with `dimension=4`, and this code will happily accept it.

mupt/geometry/measure.py
```diff
@@ -3,16 +3,22 @@
...
+    VectorN,
+    ArrayNxN,
+)
 
 
 def normalize(
-        vector : np.ndarray[Shape[Any], Numeric],
-        order  : Optional[Union[int, float, str]]=None,
-    ) -> None:
+    vector : VectorN | ArrayNxN,
```
dwhswenson: Not restricted to NxN (doesn't have to be square): any matrix (array of vectors) will work here.

mupt/geometry/shapes/shapes.py
```diff
@@ -0,0 +1,121 @@
...
+        '''Return a scaled copy of this shape'''
+        new_shape = self.copy() # works because RigidlyTransformable is also expected to be copyable
+        new_shape.scale(scaling_factor)
+        
+        return new_shape
+    
+    def __eq__(self, other : Self) -> bool:
+        return super().__eq__(other) and self.transformed_like(other)
+        
+class Shaped(Protocol):
```
dwhswenson, line 87: Naming on this is unclear. It appears that this is some sort of "thing that contains a shape," or maybe a mix-in for things that might contain a shape. This needs a little more explanation of how it is intended to be used.

mupt/geometry/shapes/pointcloud.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(self, positions : Optional[ArrayNx3]=None) -> None:
+        if positions is None:
+            positions = np.empty((0, 3), dtype=float)
+        self.positions = np.atleast_2d(positions)
+
+    def __repr__(self) -> str: 
+        return f'{self.__class__.__name__}(shape={self.positions.shape})'
+    
+    @classmethod
+    def cubic(cls, sidelen : float=1.0, centered : bool=True) -> 'PointCloud':
```
dwhswenson, line 35: What's the need for `PointCloud.cubic`? Wouldn't this be better handled as a `Cube` shape subclass? Many methods might have more efficient implementations for this special case.

mupt/geometry/shapes/cylinder.py
```diff
@@ -0,0 +1,295 @@
...
+        return f'{self.__class__.__name__}(radius={self.radius}, length={self.length})'
+    
+    @property
+    def R(self) -> float:
+        '''Alias for self.radius'''
+        return self.radius
+
+    @property
+    def L(self) -> float:
+        '''Alias for self.radius'''
```
dwhswenson:
```suggestion
        '''Alias for self.length'''
```

mupt/geometry/shapes/visualize.py
```diff
@@ -0,0 +1,55 @@
...
+
+if TYPE_CHECKING:
+    from mpl_toolkits.mplot3d import Axes3D
+
+
+def visualize_shape(
+    shape : BoundedShape,
+    ax : Optional['Axes3D']=None,
+    grid : bool=True, 
+    **kwargs,
```
dwhswenson, line 16: Docstring should document how kwargs are used (what methods to look at for usage; what order of precedence)

dwhswenson (changes requested):
Some minor changes recommended. A few thoughts:

* I'm not sure that going too deep into perfecting this typing system is going to be worth the effort.
* Not in the code covered in this PR, but it looks like `Primitive` can have a `role` attribute, which is not copied in `_copy_untransformed`. Is this intentional? I would have expected that would also be copied.
* Generally, watch out for redundant copy operations -- I caught one of them here, and it looks like you might find others if you check carefully. Those can lead to notable hits on performance.

mupt/geometry/transforms/rigid/__init__.py
timbernat: This is a holdover from `random_isometry()` in the old version of the [`mupt-shapes` example](https://github.com/MuPT-hub/mupt-examples/blob/main/examples_repr/mupt_shapes.ipynb), where the `centered` arg was the _only_ arg required. I kept it in the interest of interpretability from the user, but since it's documented in the docstring I could easily switch to accept `translation_bound` only, mitigating the contradictory value edge case you point out.

mupt/geometry/arraytypes.py
timbernat: That's a fair point, though I'm not sure how to indicate that constraint purely with type hints; is this maybe just something to document in the docstring and leave at that?

I actually amended the typing of this example to be more permissive (not just arrays) and added tests for it in a [commit](https://github.com/MuPT-hub/mupt/commit/430b30081730b5afbcc46afc5447b61fe1391ab3) on another working branch more recently; maybe I ought to cherry-pick that commit into this set of changes instead?

mupt/geometry/measure.py
timbernat: This is a typo on my part, there's already an [ArrayNxM type](https://github.com/MuPT-hub/mupt/blob/better-shapes/mupt/geometry/arraytypes.py#L62) in arraytypes which seems more appropriate here

mupt/geometry/shapes/shapes.py
timbernat: Mixin is closest, moreso a BoundedShape which can be acted upon by rigid transformations and which is implicitly copyable. The point in `__eq__` here is to enforce a stronger form of equivalence on the subtype of shapes which can be transformed, namely that instances aren't just congruent **but also** occupy the same place in space. In other words, the return type is covariant with its `BoundedShape` base, as we'd expect from substitutable types. 

What I call [BoundedShape and BoundedTransformableShape here used to be one and the same](https://github.com/MuPT-hub/mupt/blob/main/mupt/geometry/shapes.py#L77-L106), but I wanted to separate the interface-only methods (BoundedShape here) from the default implementations and assertions about copyability that come with RigidlyTransformable. The interface for just "shape-y" contracts can still be referenced, while the transformability functionality is supplied in concrete subtypes to [give Primitive what it expects viz transformations of its pieces](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L1099-L1109).

mupt/geometry/shapes/pointcloud.py
timbernat: It may well be, this came up mostly in writing tests where I wanted a cluster of points which had a well-defined and reproducible boundary+volume that I could check. It was harder to write tests for a complicated set of points (especially a stochastic one), so I wanted a convenient way to initialize PointClouds for which I could easily calculate its properties a second way to verify the ConvexHull results, to wit the vertices of a cube.

mupt/mupr/primitives.py
timbernat: The reason for the copy is to avoid the original and its clone referencing the same shape object, analogous to a deepcopy. A really insidious hidden bug I this circumvents is transforming one repeat unit to move it into place and finding all copies of it in a polymer have moved along with it in lockstep!

I see what you mean about the shape setter, although the fact that `shape.cumulative_transformation` is updated there also sort of circumvents the updating done in the RigidlyTransformble.copy() boilerplate. I don't think it'll matter though, since both calls execute within the scope of `_copy_untransformed()` here.

mupt/geometry/measure.py
timbernat: Resolved in [ed7fd0b](https://github.com/MuPT-hub/mupt/pull/55/commits/ed7fd0bec3f04d6a5018bd1474d84c52734fddf2)

mupt/geometry/transforms/rigid/__init__.py
timbernat: Resolved in [8ab9028](https://github.com/MuPT-hub/mupt/pull/55/commits/8ab9028b72a43bd65beeb42d0946ff2537309182)

mupt/mupr/primitives.py
timbernat: Resolved in [82f9c0d](https://github.com/MuPT-hub/mupt/pull/55/commits/82f9c0d14fa60c00922492eb7469c53dfb43bfbc)

mupt/geometry/arraytypes.py
timbernat: Moved changes from working branch in [bf03f6e](https://github.com/MuPT-hub/mupt/pull/55/commits/bf03f6eaab788f98c196232788159e00db519632); happy to mark resolved if those changes address what you bring up here

mupt/geometry/shapes/visualize.py
timbernat: Resolved in [e03a626](https://github.com/MuPT-hub/mupt/pull/55/commits/e03a62631c0a71513ef3848d6a5445f0671505b9)

timbernat:
@dwhswenson I made the changes you recommended and left answers to the more "discussion-y" points of feedback. When you have time, could you re-review and let me know if I've addressed all your concerns?

dwhswenson (dismissed):
Looks okay now, except that you need to fix the merge conflicts from upstream changes. I'd still lean toward moving `cubic` out of the class, but that's not blocking.

timbernat:
@dwhswenson I resolved the merge conflict, which ended up being removal of \_\_author\_\_ tag from a file that no longer exists in the branch. I left a TODO note [in this commit](https://github.com/MuPT-hub/mupt/pull/55/commits/1282f58b11e1f3a65908ce5de598b830874cf412) to add a Cubic class down the line - I don't think it's worth spending the time on urgently. Given the merge conflict is resolved and the CI passes, am I OK to override-merge and update the downstream examples branch (and [references to in in CI](https://github.com/MuPT-hub/mupt/blob/better-shapes/.github/workflows/CI.yaml#L69)) that is affected by these changes?

.github/workflows/CI.yaml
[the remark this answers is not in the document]
timbernat: Resolved in pre-merge commit [5d4758d](https://github.com/MuPT-hub/mupt/pull/55/commits/5d4758dea2dcf509eeb613ebb91dc66227fe8c27)

PR #56, opened 2026-05-01
Improving cohesion of Primitive

### LINKED ISSUE #9: Compartmentalization of info stored in Primitive
[Primitive](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py), at least as of 0.1.0a1 is currently responsible for carrying around lots of information (internal and external Connector registraties, topology info, optional Shape and Element, child Primitives, etc.) and has low [cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science)). Relatedly, Primitive is currently always mutable and requires a [litany of self-consistency checks](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L446-L520) any time a hierarchy is modified to preserve important class invariants, which increase overhead and cognitive load when reasoning about code involving Primitives.

**Proposal: split up functionality among related sub-types of Primitive**
The proposed solution is to divvy up case-specific functionality among different subtypes of a base Primitive, each supporting different pieces of information and having mutable and immutable variants (compare, e.g., with OpenFF [Molecule](https://github.com/openforcefield/openff-toolkit/blob/main/openff/toolkit/topology/molecule.py#L5158) vs. [FrozenMolecule](https://github.com/openforcefield/openff-toolkit/blob/main/openff/toolkit/topology/molecule.py#L852))

Specifically, the proposed inheritance tree will look something like:
* Primitive (BasePrimitive) - supports local connectors and shape and parent attachment
  * Simple Primitive - immutable and explicitly bans attachment of children
    * AtomicPrimitive - SimplePrimitive which stores Element info (found nowhere else)
  * CompositePrimitive - Primitives which support attaching children at initialization time (immutable)
    * MutableCompositePrimitive - mutable variant of CompositePrimitive (most like the current Primitive) supports resolution shift operations, but returns immutable CompositePrimitive instances from these

The above hierarchy abides by the [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle) and ensures that case-specific information is compartmentalized to the subtypes of Primitive which require it.

**Interrim plans**
If the above hierarchy proves to time-consuming, or is needed to be implemented more urgently, would prioritize implementing AtomicPrimitive first, as it come up quite often in the examples (and likely in other applications) and represents the finest level of resolution MuPT currently support (namely, the periodic table, which is also conveniently pre-defined)

timbernat:
## Description
Aims to address points brought up in #9.

_NB: this PR is a bumped version of the now-closed PR #15, whose source branch was split into two (this PR and PR #55) to keep content changes as atomic as possible_

## Todos
  - [x] Break down Primitive into more cohesive, substitutable hierarchy of subtypes, viz:
    - [x] Primitive (BasePrimitive) - supports local connectors and shape and parent attachment
      - [x] RootPrimitive: stores system size info and global properties about mutability etc.
        - [x] Explicitly not allowed a parent; possible entry point into exporters and other ingestors
      - [x] CompositePrimitive - Primitives which support attaching children at initialization time (immutable)
        - [x] MutableCompositePrimitive - mutable variant of CompositePrimitive (most like the current Primitive) supports resolution shift operations, but returns immutable CompositePrimitive instances from these
      - [x] Simple Primitive - immutable and explicitly bans attachment of children
        - [x] AtomicPrimitive - SimplePrimitive which stores Element info (found nowhere else)
  - [ ] Move responsibility of tracking bonds directly to Connector
    - [x] Add neighbor property and encapsulation for forming/breaking bonds with alignment
      - [ ] Implement efficient neighbor collection (e.g. for exporting cross-sections and for detecting connected components)
    - [ ] Implement "cross-sections" in Primitive which select a topology based on Primitive criterion and Connector neighbor info
  - [ ] Update RDKit interfaces (and chemical validation in .chemistry) to act specifically on AtomicPrimitives

## Questions
- [ ] How should Primitive.children vs Primitive.children_by_handle be exposed? The latter contains more useful functionality for actually referencing children, but the former is likely the first thing users will gravitate towards; overriding it may not play nicely with the treelike behavior granted by the anytree.Node inheritance

## Status
- [ ] Pass tests
  - [ ] Include new tests to cover Primitive subtype behaviors
- [ ] Verify examples still run
- [ ] Ready to go


joelaforet:
Hi @timbernat , I was snooping over the code here and was wondering what the extra shape routines will be used for? My understanding was that for this PR you were going to introduce the new Primitive types like AtomicPrimitive/CompositePrimitive, but it seems there are also around 1k LOC targeted at enhancing the geometry module. Would this be better structured as a separate PR, where one is for geometry, and then the Primitive re-design builds off of that? Pinging @mrshirts , @erjank, @StephMcCallum , and @dwhswenson for thoughts as well. The code looks good so far, I'm just suggesting breaking this up so we don't eventually have to review a single 5k LOC PR.

timbernat:
@joelaforet the geometry changes _are_ already structured as a separate PR, viz #55; this PR is written assuming those changes have already been merged, which I'm communicating verbally w/ @dwhswenson to move forward with review of just the geometry parts. As far as the why, eventually it would be useful to use rigid bodies for intersection testing to allow polymers with more complex morphologies to be aware of one another when building at CG. I also wanted to close #10 first, since it's been open (both here and in my own TODOs) for a long time and because it doesn;t share much surface area w/ other ongoing changes

PR #58, opened 2026-05-05
Refactor SAAMR topology traversal for MDAnalysis/RDKit exporters

### LINKED ISSUE #51: Refactor RDKit exporter to use role-based hierarchical tree traversal

### Architecture

### Return type rationale

### Breaking change

### Proposed scope of work

**Additional context**

### LINKED ISSUE #62: Flatten() + expand() operations in the RDKit exporter scale poorly for large polymer systems
# Problem

```python
def flatten(self) -> None:
            self.expand(child_handle)

    clone_primitive = self.copy()
    clone_primitive.flatten()
    return clone_primitive
```

```python
import logging

from functools import reduce


    mol_incs.append(
    )

```

# Proposed Solution

## Benchmark code

```python
cd mupt-examples

mamba activate mupt-env
cd mupt
pip install .
```

```bash
  --sizes all \
```

This writes:

```text
```

Image

## Runtime data

```text
```

vs

## Profiling code

```bash
  --profile-n-chains 8 \
  --profile-chain-len 12 \

  --profile-n-chains 8 \
  --profile-chain-len 12 \
```

```text
Primitive.flattened
  -> Primitive.flatten
    -> Primitive.expand
        -> Primitive.check_connectors
```

joelaforet:
## Summary
- Adds shared SAAMR role topology traversal helpers under `mupt.interfaces._shared`.
- Refactors MDAnalysis all-atom export to use the shared indexed traversal.
- The [previous implementation](https://github.com/MuPT-hub/mupt/blob/main/mupt/interfaces/mdanalysis/strategies.py#L222-L245) called DFS traversal multiple times for each "role-level". Rather than run DFS more than once, we instead populate a [dataclass ](https://github.com/MuPT-hub/mupt/blob/issue-51/shared-saamr-topology-index/mupt/interfaces/_shared/topology.py#L15-L24) to hold the information about node data corresponding to the different roles as we iterate. This means we only call [DFS once per tree](https://github.com/MuPT-hub/mupt/blob/issue-51/shared-saamr-topology-index/mupt/interfaces/_shared/topology.py#L39-L150).
- Adds shared-topology tests.

## Stack Position
This is PR 1 of 3 in the role-aware RDKit/SDF export stack.

Stack:
1. This PR: shared SAAMR topology traversal.
2. #59: role-aware RDKit Mol export.
3. #60: export-only `mupt.muptio.sdf` writer.

## After This Merges
After merging this PR into `main`:
- Retarget #59 from base `issue-51/shared-saamr-topology-index` to base `main`.
- Do not rebase #59 unless GitHub reports conflicts.
- If conflicts appear, merge updated `main` into #59's branch, resolve, and push.

## Verification
- `python -m pytest mupt/tests/interfaces/test_shared_topology.py -q`
- `python -m pytest mupt/tests/interfaces/mdanalysis/test_exporters.py -q`

### DIFF
--- a/mupt/interfaces/_shared/__init__.py
+++ b/mupt/interfaces/_shared/__init__.py
@@ -1,0 +1,4 @@
+"""Shared helpers for MuPT interface backends."""
+
+__author__ = "Joseph R. Laforet Jr."
+__email__ = "jola3134@colorado.edu"
--- a/mupt/interfaces/_shared/topology.py
+++ b/mupt/interfaces/_shared/topology.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass, field
 
 from ...chemistry.core import BOND_ORDER
 from ...mupr.embedding import ConnectorReference
-from ...mupr.primitives import Primitive, PrimitiveHandle
+from ...mupr.primitives import Primitive
 from ...roles import PrimitiveRole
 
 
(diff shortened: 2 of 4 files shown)


joelaforet (commented):
No comments on this one, happy to answer any questions/confusions.

timbernat:
Thanks for splitting #22 into these more manageable chunks for review. I'll get back to doing a more thorough review/testing early next week, but the general outline of moving SAAMR-specific but MDAnalysis non-specific functionality to an outside module seems sound; I'll provide more detailed feedback on review.

erjank (approved):
Brief scan looks like totally appropriate implementation to build off of.

timbernat (approved):
lgtm

timbernat:
Per the given post-merge instructions, #59 has had it's target redirected to the `main` branch. Keeping this branch alive for now in case of any future conflicts

PR #59, opened 2026-05-05
Add role-aware RDKit molecule export

### LINKED ISSUE #31: RDKit round-trip fails on heterocyclic aromatic compounds
**Describe the bug**
Chemical structures containing certain [heterocyclic aromatic compounds](https://en.wikipedia.org/wiki/Aromaticity#Heterocyclics), such as [thiophenes](https://en.wikipedia.org/wiki/Thiophene), [pyrroles](https://en.wikipedia.org/wiki/Pyrrole), and [maleimides](https://en.wikipedia.org/wiki/Maleimide) are unable to be read into a MuPT representation from SMILES and written back out to RDKit, owing to valence errors. Occurs as a result of heteroatoms (such as the sulfur in a thipohene ring) being given an "extra" valence by RDKit's aromaticity model (2x 1.0 order single bonds-> 2x 1.5 order aromatic bonds), which is caught by [`Primitive.check_valence()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/mupr/primitives.py#L177-L183). Causes repeat units and complete polymer systems alike to be effectively unexportable out of MuPT `Primitive` instances.

**Expected behavior**
Primitives can be instantiated from SMILES (containing aromatic heterocycles or otherwise) via `primitive_from_smiles()`and then cast to an RDKit mol object via `primitive_to_rdkit()`

**Actual behavior**
Receive ValueError on `primitive_to_rdkit(...)` call, along the lines of:
```python
>> ValueError: Atomic 2-functional Primitive "5" with total valence 3 incompatible with assigned element S
```

**To Reproduce**
Can minimally reproduce with the following snippet:
```python
from mupt.interfaces.smiles import primitive_from_smiles
from mupt.interfaces.rdkit.exporters import primitive_to_rdkit

smiles : str = '*-[C:1]1=C-C=[C:2](-S-1)-*'
smiprim = primitive_from_smiles(smiles, ensure_explicit_Hs=True)
rdmol = primitive_to_rdkit(smiprim)
```

The  value of `smiles` in the above snippet can be replaced by any of the following and still reproduce the Exception (apart from the "oxydianiline" family, which are exported without error).:
```python
heterocyclic_aromatic_smiles = {
    'polythiophene' : {
        'head_thiophene' : '[H]-[C:1]1=C-C=[C:2](-S-1)-*',
        'mid_thiophene'  : '*-[C:1]1=C-C=[C:2](-S-1)-*',
        'tail_thiophene' : '*-[C:1]1=C-C=[C:2](-S-1)-[H]',
    },
    'polypyrrole' : {
        'head_pyrrole' : '[H]-[C:1]1=C-C=[C:2](-[NH]-1)-*',
        'mid_pyrrole'  : '*-[C:1]1=C-C=[C:2](-[NH]-1)-*',
        'tail_pyrrole' : '*-[C:1]1=C-C=[C:2](-[NH]-1)-[H]',
    },
    'Kapton' : {
        'head_oxydianiline' : '[NH2]-[c:1]1ccc(cc1)-O-c2cc[c:2](cc2)-*', # these are OK - kept to complete Kapton
        'mid_oxydianiline'  : '*-[c:1]1ccc(cc1)-O-c2cc[c:2](cc2)-*',     # these are OK - kept to complete Kapton
        'tail_oxydianiline' : '*-[c:1]1ccc(cc1)-O-c2cc[c:2](cc2)-[NH2]', # these are OK - kept to complete Kapton
        'head_pyromellitimide' : '[H]-[N:1]1C(=O)c2c(C(=O)1)cc3c(c2)C(=O)[N:2](C(=O)3)-*',
        'mid_pyromellitimide'  : '*-[N:1]1C(=O)c2c(C(=O)1)cc3c(c2)C(=O)[N:2](C(=O)3)-*',
        'tail_pyromellitimide' : '*-[N:1]1C(=O)c2c(C(=O)1)cc3c(c2)C(=O)[N:2](C(=O)3)-[H]*',
    }
}
```

**Screenshots**
Attached are skeletal diagrams of the thiophene, pyrrole, and maleimide chemistries on which this bug was observed:

Image

Image

Image

**Environment**
 - OS: Ubuntu 20.04.6 LTS
 - Toolkit version/commit: 0.1.1

**Additional context**
Discovered this limitation while attempting to compile example SMILES for common polymers for the [ellipsoid builder notebook](https://github.com/MuPT-hub/mupt-examples/blob/main/examples_system/ellipsoidal_chain_placement.ipynb) in `mupt-examples`. Even when inputting explicitly kekulized rings, the bonds in the heterocycles (such as the examples presented )appear to still get marked with AROMATIC bond type when calling [`chemistry.sanitization.sanitized_mol()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/chemistry/sanitization.py#L25-L74) or it calling function [`interfaces.smiles.primitive_from_smiles()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/interfaces/smiles.py#L27-L53) with `aromaticity_model=None`; indicating that maybe the underlying implementation of the molecule sanitization needs to have null aromaticity explicitly enforced when requested.

Interestingly, also tested with pyridine-containing fragments (via [poly(4-vinylpyridine)](https://www.sigmaaldrich.com/US/en/product/aldrich/472344?srsltid=AfmBOopsT0_12cV9eA5zPNhUvVcKFslWVCy8NQmxlsvk-4nUmzC36n48), SMILES used inserted below) and did **not** encounter the same valence error. In retrospect, this makes sense, as the valence on the pyridine nitrogen is the same between the aromatized and Kekule structures.

```python
"poly(4-vinylpyridine)" : {
    "head_4-vinylpyridine" : "[H]-[CH:1](c1ccncc1)[CH2:2]-*",
    "mid_4-vinylpyridine"  : "*-[CH:1](c1ccncc1)[CH2:2]-*",
    "tail_4-vinylpyridine" : "*-[C:1](c1ccncc1)=[CH2:2]"
},
```

Image

### LINKED ISSUE #51: Refactor RDKit exporter to use role-based hierarchical tree traversal

### LINKED ISSUE #62: Flatten() + expand() operations in the RDKit exporter scale poorly for large polymer systems

joelaforet:
# Summary
- Adds strategy-based role-aware RDKit export.
- Adds `primitive_to_rdkit_mols(...)`, yielding one RDKit `Mol` per SEGMENT.
- Preserves legacy `primitive_to_rdkit(...)`.
- Adds focused RDKit exporter tests.

## Issue Coverage
Addresses #31 for the new role-aware SAAMR export path.

Issue #31 exposed a mismatch between RDKit's aromaticity model and MuPT's legacy `flattened` export path. When a heteroaromatic repeat unit such as thiophene is read from SMILES, MuPT first builds/sanitizes an RDKit molecule in [`primitive_from_smiles()`](https://github.com/MuPT-hub/mupt/blob/4c5204c43f452d0f7deba8398c9705227883282b/mupt/interfaces/smiles.py#L27-L53). During RDKit import, MuPT stores each RDKit bond type directly on a `Connector` in [`connector_between_rdatoms()`](https://github.com/MuPT-hub/mupt/blob/4c5204c43f452d0f7deba8398c9705227883282b/mupt/interfaces/rdkit/components.py#L185-L207). For thiophene sulfur, that means two `AROMATIC` connectors.

The legacy `primitive_to_rdkit(...)` exporter then flattens the hierarchy and calls `child_prim.check_valence()` before constructing the output RDKit molecule ([legacy flatten/check path](https://github.com/MuPT-hub/mupt/blob/4c5204c43f452d0f7deba8398c9705227883282b/mupt/interfaces/rdkit/exporters.py#L59-L84)). `check_valence()` compares allowed element valence against `Primitive.valence` ([implementation](https://github.com/MuPT-hub/mupt/blob/4c5204c43f452d0f7deba8398c9705227883282b/mupt/mupr/primitives.py#L191-L218)), and `Primitive.valence` is computed by summing numeric connector bond orders. Those numeric orders come from RDKit's `BondTypeAsDouble()` table ([`BOND_ORDER`](https://github.com/MuPT-hub/mupt/blob/4c5204c43f452d0f7deba8398c9705227883282b/mupt/chemistry/core.py#L23-L44)), so two aromatic sulfur bonds are summed as `1.5 + 1.5 = 3`. MuPT therefore rejected a chemically valid thiophene sulfur before RDKit could receive and validate the exported graph.

This PR does not remove `check_valence()` or claim that valence checks are unimportant. `check_valence()` remains useful as a simple MuPT-side consistency check for obvious invalid atomic primitives. The problem is that this local scalar check is not expressive enough to be the final chemistry validator for RDKit export, especially for aromatic systems where RDKit's sanitization and aromaticity model determine valid valence. For the role-aware RDKit exporter, MuPT should validate the role/topology structure, preserve the original connector bond types, and then let RDKit validate the constructed chemistry.

This PR avoids the failure mode for SAAMR export by bypassing the legacy flatten-and-check path. The new exporter validates and indexes the SAAMR role hierarchy with `build_saamr_role_topology_index()` through [`AllAtomRDKitExportStrategy`](https://github.com/MuPT-hub/mupt/blob/4c5204c43f452d0f7deba8398c9705227883282b/mupt/interfaces/rdkit/strategies.py#L76-L141), then constructs an RDKit `Mol` directly from the collected atom and connector references. Bonds are added back to RDKit with their original connector bond types in [`_add_rdkit_bonds()`](https://github.com/MuPT-hub/mupt/blob/4c5204c43f452d0f7deba8398c9705227883282b/mupt/interfaces/rdkit/exporters.py#L207-L219), and the resulting RDKit molecule is validated with RDKit's property cache ([`_mol_from_rdkit_data()`](https://github.com/MuPT-hub/mupt/blob/4c5204c43f452d0f7deba8398c9705227883282b/mupt/interfaces/rdkit/exporters.py#L270-L287)).

Regression tests now cover thiophene, pyrrole, and pyromellitimide repeat units, plus an explicit RDKit sanitization check that exported thiophene retains aromatic bonds and a valid sulfur valence ([tests](https://github.com/MuPT-hub/mupt/blob/4c5204c43f452d0f7deba8398c9705227883282b/mupt/tests/interfaces/rdkit/test_exporters.py#L152-L191)).

## Stack Position
This is PR 2 of 3 in the role-aware RDKit/SDF export stack.

Blocked by #58.

Resolves #31 

Stack:
1. #58: shared SAAMR topology traversal.
2. This PR: role-aware RDKit Mol export.
3. #60: export-only `mupt.muptio.sdf` writer.

## After #58 Merges
After #58 merges into `main`:
- Retarget this PR from base `issue-51/shared-saamr-topology-index` to base `main`.
- Do not rebase unless GitHub reports conflicts.
- If conflicts appear, merge updated `main` into this branch, resolve, and push.

## After This Merges
- Retarget #60 from base `issue-51/rdkit-role-aware-export` to base `main`.
- Do not rebase #60 unless conflicts appear.

## Verification
- `python -m pytest mupt/tests/interfaces/test_shared_topology.py mupt/tests/interfaces/rdkit/test_exporters.py -q`


mupt/interfaces/rdkit/exporters.py
```diff
@@ -132,4 +140,176 @@ def primitive_to_rdkit(
         mol.SetProp(RDMOL_NAME_WRITE_PROP, primitive.label)
     
     return mol
+
+
+PDB_CHAIN_IDS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
joelaforet, line 150: Note: this block is added so that exported systems have valid PDB Chain identifiers. Problems arise when one creates a topology from a file that has unspecified chain information. By having this "wrap around" logic, we ensure each residue has a unique chain/residue index. One chain per "segment" only allows for 26 unique segments in the topology. This messes with downstream analysis, since there are potentially multiple matches for something like "Chain A Residue 1". The wraparound gives us 26 x 9999 ~= 260,000 unique residues in the system, where a `residue` is the same concept as a `repeat_unit`. In the future, we may consider switching to PDBx format exclusively, but this works fine for now.

mupt/interfaces/rdkit/strategies.py
```diff
@@ -0,0 +1,143 @@
...
+    atom_particle_labels: list[str] = field(default_factory=list)
+    atom_resids: list[int] = field(default_factory=list)
+    bonds: list[tuple[int, int]] = field(default_factory=list)
+    bond_refs: list[
+        tuple[Primitive, tuple[ConnectorReference, ConnectorReference]]
+    ] = field(default_factory=list)
+    linker_refs: list[tuple[int, Primitive, ConnectorReference]] = field(default_factory=list)
+
+
+class RDKitExportStrategy(ABC):
```
joelaforet, line 43: Note, added for completeness in case we want to export non-AA systems to RDKit. I don't know if that is needed, and this may be overkill. Happy to refactor to remove the ABC if needed.

joelaforet (commented):
Added comments for reviewers.

timbernat:
Thanks for making the code to review here as concise as it is, I'll have a review with feedback in shortly alongside #58 and #60. The issue with the scaling of flattening is definitely a bottleneck for RDKit export, but I'm working on a more general protocol for that kind of system traversal for export in #56; I'll get back on that once I have something more concrete to demonstrate. Regarding SAAMR, I want to make sure users aren't forced into SAAMR-only systems in order to use the RDKit export. There should be flexibility with the kinds of hierarchies users can put out to RDKit (i.e. more or fewer levels, ragged hierarchies like in the examples, etc.), though the preservation of traces through the hierarchy **like** for SAAMR will always be important for provenance and even losslessly porting MUPT systems thru RDKit.

Regarding aromaticity as pertains to #31, any complete resolution of that issue is incomplete without some invocation of [RDKit aromaticity models](https://www.rdkit.org/docs/RDKit_Book.html#aromaticity), likely MDL for heterocycle chemistries from experience, to be considered fully resolved. For the sulfur example given, the root issue is that the two sulfur bonds, _even when provided as single in the SMILES_ are automatically aromatized as-is, i.e. one can't initialize them as 1 + 1 from SMILES even if one wanted to. I'd want to have tests for more chemistries (I'll put those in) and bake that awareness into the valence checker itself by providing extra allowed valences for problematic N/S atoms beyond those currently tabulated for resolution of #31. Still, I appreciate the attention given to that, and the tests drawn from that Issue here.

joelaforet:
> Regarding SAAMR, I want to make sure users aren't forced into SAAMR-only systems in order to use the RDKit export. There should be flexibility with the kinds of hierarchies users can put out to RDKit (i.e. more or fewer levels, ragged hierarchies like in the examples, etc.), though the preservation of traces through the hierarchy like for SAAMR will always be important for provenance and even losslessly porting MUPT systems thru RDKit.

That was the reason for introducing the `RDKitExportStrategy` layer rather than baking the SAAMR traversal directly into `primitive_to_rdkit_mols(...)`. The current default strategy is SAAMR/all-atom because that is the workflow I needed to get ionomers working, but the exporter itself consumes `RDKitMolData` from a strategy. A future strategy could support "ragged" hierarchies, fewer/more role levels, or a more general traversal protocol from #56 while still preserving hierarchy traces in RDKit atom props.

> Regarding aromaticity as pertains to https://github.com/MuPT-hub/mupt/issues/31, any complete resolution of that issue is incomplete without some invocation of [RDKit aromaticity models](https://www.rdkit.org/docs/RDKit_Book.html#aromaticity), likely MDL for heterocycle chemistries from experience, to be considered fully resolved.

I'm not familiar enough with the MDL aromaticity model to know which additional cases it would change here, but I'm happy to add tests for specific chemistries you think this path may still mishandle.

For this PR, my intent was not to fully solve aromaticity perception inside MuPT's own valence checker. The practical failure in #31 was that the MuPT->RDKit export path rejected valid heteroaromatic systems before RDKit had a chance to validate them. The new role-aware exporter avoids that by preserving the connector bond types and letting RDKit validate the constructed molecule.

I added [regression tests ](https://github.com/MuPT-hub/mupt/pull/59/changes#diff-69c7a04e2b9dd334fafd4efc39c2dd58d53fa8a5a1ddf260b2bb232140e9f681R175-R192) for the chemistries discussed around the issue, including thiophene, pyrrole, and maleimide/pyromellitimide-like repeat units, and those now export as expected. So I think this fixes the reported RDKit export failure, while a broader aromaticity-aware update to `check_valence()` could be handled separately if we want MuPT itself to reason about those valences more completely.

erjank (dismissed):
lgtm

mupt/interfaces/rdkit/exporters.py
```diff
@@ -132,4 +140,176 @@ def primitive_to_rdkit(
...
+def _pdb_chain_and_resid(global_residue_idx: int) -> tuple[str, int]:
+    """Return PDB-compliant chain/residue identifiers for a global residue index."""
+    chain_idx, resid_offset = divmod(global_residue_idx, PDB_MAX_RESIDUE_NUMBER)
+    if chain_idx >= len(PDB_CHAIN_IDS):
+        raise ValueError(
+            "Role-aware RDKit export exceeded PDB chain/residue capacity "
+            f"({len(PDB_CHAIN_IDS) * PDB_MAX_RESIDUE_NUMBER} residues). "
+            "Use a topology format with larger identifier fields."
+        )
+    return PDB_CHAIN_IDS[chain_idx], resid_offset + 1
```
dwhswenson, line 171: Just to be clear, this is intended to ignore the usual semantics of PDB chains, right? (Because you may have more than 26 actual "chains", or may have one chain with way more than 9999 residues?) Please add some details to the docstring to help the reader.

At least, when I look at this, it looks like `_pdb_chain_and_resid` will put the first 9999 residues in the same "chain," no matter how they're actually organized.

I'm also not seeing tests that cover how RDKit will treat these larger edge cases. You should test that the RDMol you get still has the right behavior in that case (I'm also very unsure what happens if a user exports the RDMol to PDB via the usual methods there; probably something to think about in #60 as well.

mupt/interfaces/rdkit/strategies.py
```diff
@@ -0,0 +1,143 @@
...
+            self.default_atom_position = default_atom_position
+
+    @property
+    def label(self) -> str:
+        """Human-readable strategy name."""
+        return "All-atom"
+
+    def validate(self, root: Primitive) -> None:
+        """Validate role assignments needed for all-atom RDKit export."""
+        build_saamr_role_topology_index(root)
```
dwhswenson: Do we have a case where we directly use `validate`? It seems we're just using it implicitly in `iter_mol_data` here (and `collect_topology` for MDAnalysis). All of these rely on the fact that `build_saamr_role_topology_index` will error if there's a problem, but if you ever add additional validation in `validate`, you won't validate that logic in `iter_mol_data`.

dwhswenson (changes requested):
Main required change is the docstring improvement to help readers understand the intent, and ideally some additional testing (see line comments).

A few big picture questions/changes to consider for now or the future:

* Are these intended to be used for import and export? I don't see any import methods implemented, so it is hard to know if the current code will round-trip.

* It seems to me that a lot of this might be better handled with explicit mix-in classes. For example, it looks like most of your (implemented) `validate` methods are just calls to `build_saamr_role_topology_index`, in both RDKit and MDAnalysis exporters. Why not have both inherit from an ABC/Protocol that provides `validate` and `label`?

* It might also be good for both strategies to provide a more similar interface. Currently, one is a generator that yields segments; the other builds a list of all segments. If you implement the former, it is easy to implement the latter. But it is potentially confusing if you have different interfaces for different export strategies. We can make this an issue for later if you'd prefer.

mupt/interfaces/rdkit/exporters.py
joelaforet: Yes, this is a hack to make PDB topologies render nicely in PyMol + enforce unique (chain, residue) indices for analysis. Each "Chain" in a PDB file can have a single character identifier, limiting us to the 26 letters of the alphabet,  in addition to 9999 residues that can be associated with that chain due to column spacing. This code is meant to wrap each residue of the generated system into a unique chain.

> At least, when I look at this, it looks like _pdb_chain_and_resid will put the first 9999 residues in the same "chain," no matter how they're actually organized.

That is correct, and it has worked well in practice from my tests.

joelaforet: > I'm also not seeing tests that cover how RDKit will treat these larger edge cases. You should test that the RDMol you get still has the right behavior in that case

Should I add a test case that has a 10,000 residue system? Or a single polymer chain with 10,000 residues? From my understanding the connectivity over chains, i.e. if a molecule has 3 residues and they are (A, 9998), (A, 9999), (B, 0001), doesn't matter if we split over chains. We just need a unique (ChainID, ResidueID) identifier. I can add this as a test.

mupt/interfaces/rdkit/strategies.py
joelaforet: > Do we have a case where we directly use validate?

No, I don't think the user will be calling `validate`. I think this makes sense to include as a private method that other methods call. Is using this method implicitly acceptable, or should it be reworked? I imagined this as a necessary check to run to ensure the Primitive in question is able to be exported by this strategy, and that we utilize the shared SAMMR helper.

mupt/interfaces/rdkit/exporters.py
dwhswenson: Main question is what happens when you try to export, especially to PDB or SDF. I expect the RDMol is probably okay (but a test for that would not hurt). However, I'm not sure what happens in output to other formats -- not sure if PDB (or RDKit's implementation of it) allows bonds between chains, or if will force it to break the bonds.

If it causes problems, we'll need to we warn users about it in the conversion to RDKit. (e.g., "Don't output this to PDB" in the docstring). We should also have tests for it in the SDF output (will add a review comment there).

mupt/interfaces/rdkit/strategies.py
dwhswenson: In general, Python developers tend to prefer the "easier to ask forgiveness than permission" approach to error handling, rather that "look before you leap" (https://realpython.com/python-lbyl-vs-eafp/), so raising an error on trying to iterate is more Pythonic that having a `validate` method (there are always edge cases where LBYL if good -- I don't think this is one of them.)

In some cases, a `validate` method is reasonable, but when that is expensive, you'd much rather use EAFP. You're currently counting on the `build_saamr_role_topology_index` function to also do validation. So I don't think you need `validate` here, as long as you ensure that the main validation step happens at the beginning of any use of an expensive function (which is already happening).

mupt/interfaces/rdkit/exporters.py
joelaforet: Added tests for this [here](https://github.com/MuPT-hub/mupt/pull/59/changes#diff-69c7a04e2b9dd334fafd4efc39c2dd58d53fa8a5a1ddf260b2bb232140e9f681R199-R227). It appears bonds between chains are okay!

joelaforet:
@dwhswenson 

> Are these intended to be used for import and export? I don't see any import methods implemented, so it is hard to know if the current code will round-trip.

Do you think I should add import utilities in this PR so we have round-tripping?

Also what do you think about keeping/removing [this implementation](https://github.com/MuPT-hub/mupt/pull/59/changes#diff-47d78fe89eeb602f80fda9eb25485d44d7f1262a8bb1dcd1aa625fc2d11d30f2R43-R55) of primitive_to_rdkit()? The benchmarking I did in #62 leads me to believe we should remove this implementation, but that would change the API and some prior tests.

mupt/interfaces/rdkit/exporters.py
```diff
@@ -44,7 +47,12 @@ def primitive_to_rdkit(
     '''
     Convert a Primitive hierarchy to an RDKit Mol
     Will return as single Mol instance, even is underlying Primitive represents a collection of multiple disconnected molecules
-    
+
+    DEV: This is the legacy flattened exporter. We recommend replacing downstream
```
joelaforet, line 50: Mark as deprecated, make a separate PR that actually removes this.

joelaforet: Use standard deprecation warning, in warning.warn

dwhswenson (approved):
LGTM!

Issue #48, opened 2026-03-25
Implement versioned serialization for Primitive hierarchies


joelaforet:
## Is your feature request related to a problem? Please describe.

MuPT still does not provide a native way to save and reload a `Primitive` representation. This is a real workflow bottleneck for systems that take meaningful effort to build, especially SAAMR-compliant systems used in `mupt-examples`.

A common workflow is:
`build locally -> transfer to HPC -> run simulation -> retrieve trajectory -> analyze later`

Today, if the original Python session is gone, the MuPT representation is gone as well. That means users lose the tree-like relational hierarchy of `Primitive` objects and their children, including the information needed to recover MuPT-native analysis pathways such as `primitive_to_mdanalysis()`.

This issue is intended as a concrete implementation-oriented continuation of #11, which discussed canonical forms and serialization at a broader level. The goal here is to define and implement a practical, versioned serialization/deserialization path for `Primitive`-based representations.

## Describe the solution you'd like

Add a versioned serialization/deserialization protocol for `Primitive` hierarchies.

The main requirement is faithful preservation of the relational structure of a MuPT representation, including:
- parent/child hierarchy
- child handles / labels
- connector information
- internal and external connector mappings
- topology / graph connectivity
- geometry / shape information
- element assignments
- relevant metadata

Initial support can focus on SAAMR-compliant systems, but the design should ideally extend to more general `Primitive` trees.

A simple public API would be useful, for example:
- `save(primitive, path)`
- `load(path) -> Primitive`

The on-disk format should be versioned so future schema evolution is explicit and old files can be handled gracefully.

A schema-driven implementation is likely the safest path. In particular, a Pydantic-based schema layer is worth serious consideration, not as a replacement for MuPT core runtime classes, but as a typed serialization boundary:
- MuPT runtime objects (`Primitive`, `Connector`, `TopologicalStructure`, shapes, registries) remain unchanged
- dedicated serialization DTOs define the persisted structure
- MuPT objects are converted to DTOs on save, and DTOs are converted back to MuPT objects on load

This approach is attractive because MuPT needs a rigorously typed, recursive, versionable schema for tree-structured molecular representations. Pydantic is well suited to nested validation, recursive models, explicit serialization rules, and JSON Schema generation. JSON Schema would also give maintainers a concrete contract for `.mupt` contents.

[...]

PR #60, opened 2026-05-05
Add MuPT SDF utils (export and import)

### LINKED ISSUE #51: Refactor RDKit exporter to use role-based hierarchical tree traversal

### LINKED ISSUE #62: Flatten() + expand() operations in the RDKit exporter scale poorly for large polymer systems

joelaforet:
## Summary
- Adds export-only `mupt.muptio.sdf`.
- Adds streaming multi-record SDF writing for role-aware MuPT hierarchies.
- Re-exports `write_primitive_to_mupt_sdf(...)` from `mupt.interfaces.rdkit` for compatibility.
- Does not add SDF deserialization or claim SDF is canonical MuPT persistence.

## Stack Position
This is PR 3 of 3 in the role-aware RDKit/SDF export stack.

Blocked by #59.

Stack:
1. #58: shared SAAMR topology traversal.
2. #59: role-aware RDKit Mol export.
3. This PR: export-only `mupt.muptio.sdf` writer.

## After #59 Merges
After #59 merges into `main`:
- Retarget this PR from base `issue-51/rdkit-role-aware-export` to base `main`.
- Do not rebase unless GitHub reports conflicts.
- If conflicts appear, merge updated `main` into this branch, resolve, and push.

# Implements RDKit functionality of #22 

## Verification
- `python -m pytest mupt/tests/interfaces/test_shared_topology.py mupt/tests/interfaces/rdkit/test_exporters.py mupt/tests/muptio/test_sdf.py -q`


mupt/interfaces/rdkit/exporters.py
```diff
@@ -144,6 +144,17 @@ def primitive_to_rdkit(
 
 PDB_CHAIN_IDS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 PDB_MAX_RESIDUE_NUMBER = 9999
+MUPT_RDKIT_ATOM_PROPS = (
```
joelaforet, line 150: Added so we can inject this as metadata in the exported .sdf file.

joelaforet (commented):
Added comments for reviewers.

timbernat:
Poking to give proof of life and keep this on my review radar over the next few business days

> Does not add SDF deserialization or claim SDF is canonical MuPT persistence.

This is nice, prevents painting ourselves into a corner when adding this in the future!

A couple clarifying questions just to make sure I understand this PRs purpose:
* What do you mean by " streaming multi-record SDF writing"?
* What's the rationale for having the SDF writing directly within MuPT, given that it is (as it should be) essentially a thin wrapper around RDKit's SDF utilities? A possibility worth considering, given the surrounding RDKit writer changes, would be to have the role records and other atom info bundled into atom Primitives' metadata and routed indirectly into Props on exported RDKit Mols, where those are then read by RDKit's SDWriter "outside the fence" of MuPT.

joelaforet:
> What do you mean by " streaming multi-record SDF writing"?

I mean that the writer emits one SDF record per MuPT segment/chain and writes each RDKit `Mol` immediately as it is generated. It does not first flatten the full hierarchy into a monolithic RDKit molecule, and it does not materialize a list of all segment mols before writing (which is potentially very memory intense for a system with many polymer chains). `primitive__to_rdkit_mols(...)` is an iterator, and `write_primitive_to_sdf(...)` consumes one `Mol`, prepares its SDF atom-property lists, writes it with `SDWriter`, and then moves on, freeing up memory.

> What's the rationale for having the SDF writing directly within MuPT, given that it is (as it should be) essentially a thin wrapper around RDKit's SDF utilities? 

The rationale is mainly scalability and preserving Topology. The previous workflow flattened the MuPT Universe, exported one disconnected RDKit molecule, then used `Chem.GetMolFrags(..., asMols=True)` to rediscover the chain records. That became problematic for large systems, especially when I was testing the 800-chain ionomer system. Issue #62 has some benchmarking details for both runtime and memory usage. 

> A possibility worth considering, given the surrounding RDKit writer changes, would be to have the role records and other atom info bundled into atom Primitives' metadata and routed indirectly into Props on exported RDKit Mols, where those are then read by RDKit's SDWriter "outside the fence" of MuPT.

I'm not sure I understand this completely, but I think the current implementation is close to the spirit of this suggestion. This approach derives the role metadata at export time rather than storing it permanently in atom `Primitive.metadata`.
The role-aware RDKit exporter already attaches MuPT-specific atom props such as `chain_id`, `residue_id`, `mupt_segment_index`, `mupt_residue_index`, and particle labels to the exported RDKit atoms. The SDF helper then only performs the SDF-specific finalization step: converting those existing RDKit atom props into SDF atom-property lists with `CreateAtomStringPropertyList(...)`, then handing the mols to RDKit's `SDWriter`.

I avoided pushing these fields back into atom `Primitive.metadata` because several of them are export-context labels rather than intrinsic atom properties. For example, segment index, residue index, and SDF/PDB-compatible chain/residue identifiers are derived from the role-aware traversal and from the chosen export representation. Keeping them in the export layer avoids maintaining exporter-derived metadata inside the MuPT Representation.

This also leaves the metadata derivation strategy-dependent. The current `AllAtomRDKitExportStrategy` derives segment/residue/particle metadata in the way that makes sense for SAAMR systems, but a different `RDKitExportStrategy` could choose different record boundaries or derive different RDKit atom props.

Users who want to stay outside `mupt.muptio.sdf` can still consume `primitive_to_rdkit_mols(...)` directly and call `SDWriter` themselves. The MuPT SDF helper is mainly a convenience layer that preserves MuPT-specific record boundaries and performs the required atom-property-list preparation consistently.

timbernat:
> I mean that the writer emits one SDF record per MuPT segment/chain and writes each RDKit Mol immediately as it is generated

That makes sense, I wanted to make "serialize" here was referred to a buffer and not used in some other sense (i.e. we've used "serialize" to refer to the order of parts of a system in other discussions). 


> The previous workflow flattened the MuPT Universe, exported one disconnected RDKit molecule, then used Chem.GetMolFrags(..., asMols=True) to rediscover the chain records. That became problematic for large systems, especially when I was testing the 800-chain ionomer system. 

This is valid, and we definitely want to get away from flattening as a mechanism for export. There's nothing about RDKit's Mol, or indeed a general graph, which says they _need_ to be one connected component, but I recognize the need to be able to buffer parts of a large system in implementation for curtailing memory requirements.

For disclosure, the choice to write as a single monolithic Mol (as in [`primitive_to_rdkit()`](https://github.com/MuPT-hub/mupt/blob/main/mupt/interfaces/rdkit/exporters.py#L40)) was to enable storing system-wide information, such as box vectors or info about metadata formats and [provenance](https://github.com/MuPT-hub/mupt/blob/main/mupt/interfaces/rdkit/exporters.py#L121), within the outputted SDF. SDFs support header and non-structural data on a per-molecule basis but not on a per-file one, nor do other [CTab-based file formats](https://discover.3ds.com/sites/default/files/2020-08/biovia_ctfileformats_2020.pdf), so enforcing "always one molecule per file" was a way to get around that. This is not to say that's a _great_ way of achieving that result, but more context for why that design choice was made initially. 


> I avoided pushing these fields back into atom Primitive.metadata because several of them are export-context labels rather than intrinsic atom properties.

I see the case for that, although metadata is intended as a sort of "dumping ground" for additional arbitrary data without a dedicated place in the repr, so it's not a crime to put index fields there. It looks like you have [dedicated dataclasses elsewhere](https://github.com/MuPT-hub/mupt/blob/9fea58598ca85edff121c90113d6243f60a1e2cb/mupt/interfaces/rdkit/strategies.py#L24-L40) in this PR which bundle the same info, which works too.


> The SDF helper then only performs the SDF-specific finalization step: converting those existing RDKit atom props into SDF atom-property lists

When I said "have the role records and other atom info bundled into atom Primitives' metadata and routed indirectly into Props on exported RDKit Mols", this is more along the lines of what I was referring to. Specifically, RDKit already takes Props assigned to Mol objects and packages them into SDF metadata, which you could assign per-atom on Mol objects like you did [here](https://github.com/MuPT-hub/mupt/blob/issue-51/muptio-sdf-export/mupt/muptio/sdf.py#L29-L32) in, say `.interfaces.rdkit.exporters`, without requiring the [internal SDWriter wrapper](https://github.com/MuPT-hub/mupt/blob/issue-51/muptio-sdf-export/mupt/muptio/sdf.py#L35-L62), which cuts you off from the [configuration of chemical preferences in SDWriter](https://www.rdkit.org/docs/cppapi/classRDKit_1_1SDWriter.html) (e.g. Kekulization, V3K fmt, etc.), though I appreciate you using the [asstrpath](https://github.com/MuPT-hub/mupt/blob/issue-51/muptio-sdf-export/mupt/mutils/filepaths/pathutils.py#L65-L69) util instead of blindly using strings for Paths as people often do.

My reason for concern is that RDKit support a [slew of potential export formats](https://www.rdkit.org/docs/source/rdkit.Chem.rdmolfiles.html) out of Mol, and if we choose to specifically provide a thin wrapper for SDF writing it may set the precedent of having wrapper for _all_ these endpoints in MuPT, which is a nightmare for maintenance and adds a bunch of extra code we shouldn't have to maintain ourselves. 
**However, not all those formats support arbitrary metadata, and conversely not all SDFs with metadata have a valid interpretation, so if the intent is instead to use SDF an storage format with particular choices made about SAAMR metadata (ignoring for a second that the impl is via RDKit), then that makes more sense scope-wise.**


> Users who want to stay outside mupt.muptio.sdf can still consume primitive_to_rdkit_mols(...) directly and call SDWriter themselves. The MuPT SDF helper is mainly a convenience layer that preserves MuPT-specific record boundaries and performs the required atom-property-list preparation consistently.

For the sake of clarifying where this fits in, is it fair to describe the purpose of SDF writing here as an intermediate means of preserving SAAMR metadata in a static file format (e.g. for use in simulation setup or analysis) while a more permanent MuPT-specific file save format is decided upon and implemented? With the intent not being to able to completely recover the originating MuPT repr hierarchy from that file, but to rather to preserve PDB-style information which would otherwise be lost on flatten.

joelaforet:
> For the sake of clarifying where this fits in, is it fair to describe the purpose of SDF writing here as an intermediate means of preserving SAAMR metadata in a static file format (e.g. for use in simulation setup or analysis) while a more permanent MuPT-specific file save format is decided upon and implemented?

Yes. Once #59 and #60 make it into the source, I will write a PR that write something along the lines of a `foo.mupt.sdf` file as a temporary workaround addressing #48 , while we wait for the `foo.mupt` file format to be designed. If it makes more sense, I can add that to this PR instead.

mupt/tests/muptio/test_sdf.py
```diff
@@ -0,0 +1,69 @@
...
+
+from mupt.interfaces.rdkit import write_primitive_to_mupt_sdf
+from mupt.muptio.sdf import MUPT_SDF_ATOM_PROPS, write_primitive_to_sdf
+
+
+def _load_sdf(path):
+    return [mol for mol in SDMolSupplier(str(path), removeHs=False, sanitize=False) if mol is not None]
+
+
+def test_write_primitive_to_sdf_writes_one_record_per_segment(
```
dwhswenson: Add some more testing to this to cover edge cases where the chains need to be relabelled, as mentioned in https://github.com/MuPT-hub/mupt/pull/59#discussion_r3376184212.

mupt/temporary/sdf.py
dwhswenson, line 1: Since the intent is to treat this as a temporary format (essentially deprecated from day 1), and to delete it before a 1.0 release, it would probably be best to move this somewhere in the filesystem that indicates that (e.g., `mupt/temporary/sdf.py`). Also move the test file to a matching location. This makes it easier to ensure that everything is removed before the release.

dwhswenson (changes requested):

mupt/temporary/sdf.py
joelaforet: Files have been moved to a temporary directory!

mupt/tests/muptio/test_sdf.py
joelaforet: Added some more tests [here](https://github.com/MuPT-hub/mupt/pull/60/changes#diff-0ef63569ea8a35157cf274d49c05385e695cc54172a9b8fa9807005866f48271R152-R169).

joelaforet:
Test via a 1.5 trip, serialize, deserialize, reserialize. Write to file, read from file, write to file again.

joelaforet:
Per @dwhswenson , I will add in an import function to this code too, and then also change the file extension to be ".mupt.sdf", not just ".sdf".

joelaforet:
## MuPT SDF Read/Write Performance Notes

I ran a benchmark after the streaming importer, SDF connector reconstruction, forward-only `ForwardSDMolSupplier`, atomic writes, and per-record atom-position caching changes.

### Small/Medium Systems

These use valid SAAMR polyethylene `.mupt.sdf` files generated from the test fixtures. The read benchmarks exclude polymer construction time.

| system | mode | time | peak RSS |
|---:|---|---:|---:|
| 7k atoms / 500 records | streamed segment import | 2.57 s | 161 MB |
| 28k atoms / 2000 records | streamed segment import | 10.43 s | 163 MB |
| 28k atoms / 2000 records | materialized universe import | 11.26 s | 504 MB |

For comparison, before position caching the 28k streamed import was ~18.15 s, so caching conformer positions once per SDF record gave a meaningful speedup.

### Large-System Scale Check

To test near the largest system size we are currently considering, I generated a ~420k atom `.mupt.sdf` by concatenating valid 28k-atom MuPT SDF templates. This isolates SDF read/write scaling from polymer/random-walk build cost.

| operation | records | atoms | bonds | time | peak RSS |
|---|---:|---:|---:|---:|---:|
| streaming read | 30,000 | 420,000 | 390,000 | 162.92 s | 163.9 MB |
| materialized read | 30,000 | 420,000 | n/a | 195.73 s | 5438.2 MB |
| materialized read + rewrite | 30,000 | 420,000 | n/a | read 194.33 s, write 24.84 s | 5672.0 MB |

Takeaways:

- The streaming API (`iter_primitives_from_mupt_sdf(...)`) has effectively flat memory for many-record systems because it yields one SEGMENT at a time.
- Full `UNIVERSE` materialization works at ~420k atoms on this machine, but costs ~5.4 GB RAM.
- Writing is relatively fast once the MuPT hierarchy is already materialized (~25 s for ~420k atoms).
- The bottleneck is object/connector reconstruction, not raw SDF file I/O.

### Profiling Hotspots

I profiled streamed import on the 28k atom / 2000 record benchmark before the final position-cache optimization. The main cumulative hotspots were:

| hotspot | cumulative time | notes |
|---|---:|---|
| `_build_segment_from_mol(...)` | ~32.8 s | total per-record hierarchy rebuild |
| `Primitive.attach_child(...)` | ~12.4 s | MuPT tree attachment and child bookkeeping |
| `Primitive.bind_external_connector(...)` | ~11.2 s | connector propagation during attachment |
| `_connector_from_sdf_bond(...)` | ~10.2 s | directed connector reconstruction |
| `_atom_position(...)` / repeated `np.array(...)` | ~9.1 s | fixed by per-record position caching |
| `Connector._copy_untransformed(...)` / `deepcopy` paths | ~8.6 s | connector copies during propagation |

The obvious low-risk optimization was per-record atom-position caching, which we added. The remaining major cost is connector registration/copying/propagation in core MuPT object construction. That seems better addressed in the separate connector-handling PR rather than over-optimizing this temporary SDF bridge.

### Practical Guidance

- Use `iter_primitives_from_mupt_sdf(...)` for large workflows that can operate segment-by-segment.
- Use `primitive_from_mupt_sdf(...)` only when a full materialized `UNIVERSE` is required.
- For ~400k atom systems, the optimized reader/writer appears viable on this machine, but full materialization should be expected to require several GB of RAM.

mupt/interfaces/rdkit/__init__.py
```diff
@@ -55,6 +52,44 @@
...
+
+def write_primitive_to_mupt_sdf(
+    primitive,
+    path,
+    resname_map,
+    default_atom_position=None,
+    strategy=None,
+):
+    """Compatibility wrapper for the temporary MuPT SDF writer alias."""
+    from ...temporary.sdf import write_primitive_to_mupt_sdf as _write_primitive_to_mupt_sdf
```
dwhswenson: (not blocking) Curious why this import is local to the method? I would think that having it at module level would be quicker to notify you when `temporary.sdf` gets removed, and is also more standard.

mupt/temporary/sdf.py
```diff
@@ -1,35 +1,73 @@
...
+
+def _mupt_sdf_path(path: str | Path) -> Path:
+    """Return ``path`` normalized to the temporary MuPT SDF suffix."""
+    path = Path(path)
+    path_str = asstrpath(path)
+    if path_str.endswith(MUPT_SDF_SUFFIX):
+        return path
+    if path.suffix == ".sdf":
+        return path.with_suffix(MUPT_SDF_SUFFIX)
+    return Path(f"{path_str}{MUPT_SDF_SUFFIX}")
```
dwhswenson: You should emit a `UserWarning` if you're modifying filenames given by the user. Changing filenames without warning can cause confusion.

mupt/temporary/sdf.py
```diff
@@ -45,21 +83,370 @@ def write_primitive_to_sdf(
     generated RDKit Mol at a time so large systems do not retain every segment
     molecule in memory.
     """
+    target_path = _mupt_sdf_path(path)
+    temp_fd, temp_name = tempfile.mkstemp(
+        prefix=f".{target_path.name}.",
+        suffix=".tmp",
+        dir=asstrpath(target_path.parent),
+    )
+    os.close(temp_fd)
```
dwhswenson, line 100: Please add a comment to explain that you're using a tempfile here to ensure "atomic" writes; i.e., that you don't leave the file partially written.

I'm not sure that this is necessarily needed in this case, but comment the pattern since it won't be obvious to everyone why this is being done.

mrshirts:
@dwhswenson you had asked about adding an import functionality, which could be added, can you re-review?  @timbernat any comments?

dwhswenson (changes requested):
A couple of small changes.

Additionally, tt looks like a lot of the stuff to facilitate round-tripping is actually RDKit -> MuPT, not SDF-specific. However, it is in `temporary/sdf.py`. Is this intended to be removed when the SDF stuff is removed, or should this be moved somewhere where it will be available long-term?

joelaforet:
> Additionally, tt looks like a lot of the stuff to facilitate round-tripping is actually RDKit -> MuPT, not SDF-specific. However, it is in temporary/sdf.py. Is this intended to be removed when the SDF stuff is removed, or should this be moved somewhere where it will be available long-term?

@dwhswenson TBH I don't have a good sense of this. My understanding is that the existing RDKit module is meant to facilitate MuPT -> RDKit molecules specifically, not necessarily the SDF writing in this PR, even though we use RDKit to accomplish this. I don't know where the best place to put this is, since IMO this is a genuinely useful feature, but it doesn't apply to non SAAMR hiearchies. I think the idea is that when we design the final ".mupt" file format, we won't have any need for writing to a ".sdf", so that's why this is in a temporary module. I also think that ideas here will be reused in the ".mupt" file PR, whenever that comes about.

mupt/temporary/sdf.py
joelaforet: Comment added!

mupt/temporary/sdf.py
joelaforet: Warning added!

mupt/interfaces/rdkit/__init__.py
joelaforet: I restructured the imports, thanks for the heads up!

dwhswenson (dismissed):
Changes LGTM, although you need to fix merge conflicts and (hopefully) see tests pass. Tests may be failing due to other issues currently, however.

dwhswenson (approved):

PR #83, opened 2026-06-10
Add all-atom DPD melt initializer

joelaforet:
## Description
Adds an all-atom DPD coordinate builder for SAAMR-compliant MuPT hierarchies. The builder initializes dense, all-atom melt coordinates, sets the positions of the "leaf Primitives" (atoms) in place, and records a cubic box/summary on the root primitive.

This is solely for initialization; we claim that the AA-DPD builder will generate a set of coordinates for your polymer system at the target density that CAN BE minimized with an MD force field. We do not do the minimization for you. Running downstream minimization/MD is expected before analyzing the system.

## Todos
Notable points that this PR has either accomplished or will accomplish:
  - [x] Add `AllAtomDPDBuilder` and settings/result dataclasses
  - [x] Use the existing `PlacementGenerator` abstraction for pre-HOOMD frame-0 residue placement
  - [x] Use OpenFF labels for bonded restraints and vdW-informed DPD particle typing
  - [x] Preserve OpenFF multi-term proper/improper torsions
  - [x] Keep optional dependencies lazy (`hoomd`, `gsd`, `freud`, `openff`, `openmm`)
  - [x] Add focused builder utility tests
  - [x] Add manual dense PE AA-DPD/OpenMM validation harness
  - [x] Validate PE melt handoff with minimization, NVT, and NPT runs

## Questions
- [ ] Should the PE validation script stay under `devtools/scripts`, or move later into examples once formal docs/examples are ready?
- [ ] Are the current default AA-DPD settings acceptable as initializer defaults, or should they be more conservative?

## Sample API call
```python
from mupt.builders.all_atom_dpd import AllAtomDPDBuilder, AllAtomDPDSettings
from mupt.roles import assign_SAAMR_roles

assign_SAAMR_roles(root)

settings = AllAtomDPDSettings(
    density_g_cm3=0.77,
    initial_residue_spacing_a=5.0,
    initial_angle_max_rad=0.7853981633974483,
    n_steps_max=50000,
    n_steps_per_interval=1000,
    random_seed=42,
    resname_map=resname_map,
)

result = AllAtomDPDBuilder(settings=settings).build(root)

assert result.converged
print(result.box_length_a)
print(root.metadata["all_atom_dpd_summary"])
```

By default, AA-DPD uses `AngleConstrainedRandomWalk` as the frame-0 `PlacementGenerator` for residue placement within each segment, then runs all-atom DPD relaxation and writes final atom coordinates back to leaf `Primitive.shape` values as `PointCloud`s.

## Verification
- [x] `python -m pytest mupt/tests/builders/test_all_atom_dpd.py -q`
- [x] `25 passed`
- [x] `python -m pytest mupt/tests/builders -q`
- [x] `25 passed`
- [x] `python -m py_compile mupt/builders/all_atom_dpd.py mupt/builders/random_walk.py`
- [x] `python -m py_compile devtools/scripts/validate_all_atom_dpd_pe_melt.py`

Manual validation used dense polyethylene systems built from explicit-H PE repeat units, OpenFF `openff-2.2.1.offxml`, and NAGL/AshGC charges via `openff-gnn-am1bcc-1.0.0.pt`.

Validation cases run:
- `10 x 15` PE chains, `0.77 g/cm^3`, `450 K`, `2 fs`, `1 / ps`: AA-DPD converged, OpenMM minimization completed, `10 ps` NVT + `20 ps` NPT stayed finite.
- `15 x 20` PE chains, `0.77 g/cm^3`, `450 K`, `2 fs`, `1 / ps`: AA-DPD converged, OpenMM minimization completed, `10 ps` NVT + `20 ps` NPT stayed finite.
- `20 x 30` PE chains, `0.77 g/cm^3`, `450 K`, `2 fs`, `1 / ps`: AA-DPD converged, OpenMM minimization completed, `10 ps` NVT + `20 ps` NPT stayed finite.
- Earlier validation also covered `20 x 30` PE chains at `450 K` for `200 ps` NVT + `1 ns` NPT before the PlacementGenerator refactor.

The smallest attempted `5 x 8` PE case converged in AA-DPD but was too small for the default OpenMM nonbonded cutoff, so it was not counted as an OpenMM validation case.

## Status
- [x] Pass focused unit tests
- [x] Pass dense PE OpenMM validation at melt-like conditions
- [ ] Obtain review
- [ ] Ready to go


devtools/scripts/validate_all_atom_dpd_pe_melt.py
joelaforet, line 1: Not sure if we want to keep this in the main repo or not, this was the script I used while testing that the AA-DPD can produce minimizeable coordinates.

mupt/tests/builders/test_all_atom_dpd.py
joelaforet, line 1: Looking for suggestions on what tests to add. Should I add in a test that actually runs DPD on a real system + asserts the minimized energy is finite? That seems like a large test (heavy dependencies + compute needed to run DPD)

mupt/builders/all_atom_dpd.py
joelaforet, line 1: This implementation is a draft; I am open to discussing re-structing of architecture + ways of making the implementation more pythonic. Pinging @dwhswenson for thoughts on this!

joelaforet (commented):
Draft 1 of AA-DPD initializer. I am open to architectural feedback!

joelaforet:
Each bond/angle/torsion should be no more than a couple kcal/mol if you divide everything out. @mrshirts pinging you to comment on a more rigorous method of validating that the DPD initialized structures are reasonable.

joelaforet:
Updated the AA-DPD frame-0 initialization to use MuPT's existing `PlacementGenerator` abstraction instead of the custom linear initializer.

Implementation notes:
- Default frame-0 placement now uses `AngleConstrainedRandomWalk` per segment.
- `AllAtomDPDBuilder` still owns the all-atom DPD relaxation and writes final atom coordinates directly to leaf primitives.
- Added validation of placement-generator output handles, deterministic RNG plumbing, and a single-residue segment path.
- Reviewer pass found no blocking issues after the fixes.

Fast validation:
- `/home/joelaforet/miniconda3/envs/mupt-env/bin/python -m pytest mupt/tests/builders/test_all_atom_dpd.py -q` -> `24 passed`
- `/home/joelaforet/miniconda3/envs/mupt-env/bin/python -m pytest mupt/tests/builders -q` -> `24 passed`
- `/home/joelaforet/miniconda3/envs/mupt-env/bin/python -m py_compile mupt/builders/all_atom_dpd.py mupt/builders/random_walk.py` -> passed

Full-stack PE validation used `/home/joelaforet/miniconda3/envs/mupt-dpd-ionomer/bin/python`, explicit-H PE, `0.77 g/cm^3`, `450 K`, OpenFF `openff-2.2.1.offxml`, and NAGL/AshGC charges via `openff-gnn-am1bcc-1.0.0.pt`.

Validation matrix:
- `10 x 15` PE: AA-DPD converged in `1000` steps; OpenMM minimized to finite PE `1266.727501 kJ/mol`; `10 ps` NVT + `20 ps` NPT stayed finite.
- `15 x 20` PE: AA-DPD converged in `1000` steps; OpenMM minimized to finite PE `2744.276509 kJ/mol`; `10 ps` NVT + `20 ps` NPT stayed finite.
- `20 x 30` PE: AA-DPD converged in `1000` steps; OpenMM minimized to finite PE `6865.288969 kJ/mol`; `10 ps` NVT + `20 ps` NPT stayed finite.

joelaforet:
Follow-up clarity/adapter updates pushed locally and validated:

- Renamed the frame-0 spacing setting from `initial_bond_length_a` to `initial_residue_spacing_a` so the API describes residue PlacementGenerator spacing rather than an atomistic bond length.
- Removed redundant PE pre-placement from the validation harness; AA-DPD now owns frame-0 placement through `PlacementGenerator`.
- Preserved support for transparent hierarchy levels between canonical SAAMR roles by adapting role-aware residues into a temporary direct-child segment before calling the role-agnostic `PlacementGenerator`.
- Added a regression test for mixed direct RESIDUE + transparent grouping-node layouts to ensure residue transforms are written back to the correct original atom leaves.
- Updated validation diagnostics to report minimum-image periodic atom distances.

Validation after these updates:
- `/home/joelaforet/miniconda3/envs/mupt-env/bin/python -m pytest mupt/tests/builders/test_all_atom_dpd.py -q` -> `25 passed`
- `/home/joelaforet/miniconda3/envs/mupt-env/bin/python -m pytest mupt/tests/builders -q` -> `25 passed`
- `/home/joelaforet/miniconda3/envs/mupt-env/bin/python -m py_compile mupt/builders/all_atom_dpd.py mupt/builders/random_walk.py` -> passed
- `10 x 15` PE at `0.77 g/cm^3`, `450 K`: AA-DPD converged in `1000` steps; OpenMM minimized to finite PE `1121.280738 kJ/mol`; `10 ps` NVT + `20 ps` NPT stayed finite.

mupt/builders/all_atom_dpd.py
```diff
@@ -0,0 +1,1218 @@
...
+        parameters: _ParameterTables,
+    ) -> Any:
+        """Create and configure a HOOMD simulation from the initial frame."""
+
+        integrator = hoomd.md.Integrator(dt=self.settings.dt)
+        integrator.methods.append(hoomd.md.methods.ConstantVolume(filter=hoomd.filter.All()))
+        if bonds:
+            harmonic = hoomd.md.bond.Harmonic()
+            for name in frame.bonds.types:
+                harmonic.params[name] = parameters.bond_params.get(name, {"r0": 1.5, "k": 100.0})
```
StephMcCallum: Instead of using hard-coded defaults here for bond/angles/dihedrals/torsions of types that aren't defined by OpenFF, I suggest we set these to match another parameter set in the system, like the last parameter set used or an average of the ones in the system. 

We should also raise a warning that tells the user, 'this type' is missing and 'these parameters' were assigned.

mupt/builders/all_atom_dpd.py
```diff
@@ -0,0 +1,1218 @@
...
+        converged = self._spacing_ok(simulation.state.get_snapshot(), freud, box_lengths, excluded)
+        while not converged and steps < self.settings.n_steps_max:
+            simulation.run(self.settings.n_steps_per_interval)
+            steps += self.settings.n_steps_per_interval
+            if steps % self.settings.report_interval == 0:
+                LOGGER.debug("Integrated %s all-atom DPD steps", steps)
+            converged = self._spacing_ok(simulation.state.get_snapshot(), freud, box_lengths, excluded)
+        return steps, time.perf_counter() - start, converged
+
+    def _spacing_ok(self, snapshot: Any, freud: Any, box_lengths: np.ndarray, excluded: set[tuple[int, int]]) -> bool:
```
joelaforet, line 1471: Check the RDF to ensure no particles are below cutoff distance.

erjank:
MDtraj is a dependency to add to env

erjank:
On first-run through with default parameters, last cell of DPD_Role_Aware_SAAMR_All_Atom_Ionomer_OpenMM, running into a severe contact issue:


RuntimeError                              Traceback (most recent call last)
Cell In[13], line 400
    396         print(f"{built.system_name}: periodic contact preflight under {clash_summary['cutoff_a']:.1f} A = {clash_summary}")
    397         severe_contact = clash_summary["severe_contacts"] > 0
    398         sodium_non_oxygen_contact = clash_summary["sodium_non_oxygen_contacts"] > 0
    399         if severe_contact or (OPENMM_PREFLIGHT_ABORT_ON_NA_NON_O_CONTACT and sodium_non_oxygen_contact):
--> 400             raise RuntimeError(
    401                 f"{built.system_name}: periodic contact preflight found severe contacts before OpenMM dynamics: "
    402                 f"{clash_summary}. Na-O contacts above {OPENMM_PREFLIGHT_ALLOWED_NA_O_CONTACT_A:.2f} A are treated as "
    403                 "ion coordination rather than clashes."

RuntimeError: m5: periodic contact preflight found severe contacts before OpenMM dynamics: {'cutoff_a': 2.0, 'raw_pairs': 153473, 'nonexcluded': 1, 'sodium_contacts': 1, 'intramolecular': 0, 'intermolecular': 1, 'min_distance_a': 1.900142955148375, 'severe_contacts': 0, 'sodium_oxygen_contacts': 0, 'sodium_non_oxygen_contacts': 1, 'allowed_sodium_oxygen_contacts': 0, 'details': [{'distance_a': 1.900142955148375, 'sodium_contact': True, 'sodium_oxygen_contact': False, 'intramolecular': False, 'atom_1': 'Na1', 'atom_2': 'Na1', 'residue_1': 'SOD', 'residue_2': 'SOD', 'chain_1': 'A', 'chain_2': 'A', 'residue_number_1': '7103', 'residue_number_2': '7856'}]}. Na-O contacts above 1.80 A are treated as ion coordination rather than clashes.

mrshirts:
> MDtraj is a dependency to add to env

Can we avoid another analysis package?  What is this used for?

mrshirts:
> On first-run through with default parameters, last cell of DPD_Role_Aware_SAAMR_All_Atom_Ionomer_OpenMM, running into a severe contact issue:
> 
> RuntimeError Traceback (most recent call last) Cell In[13], line 400 396 print(f"{built.system_name}: periodic contact preflight under {clash_summary['cutoff_a']:.1f} A = {clash_summary}") 397 severe_contact = clash_summary["severe_contacts"] > 0 398 sodium_non_oxygen_contact = clash_summary["sodium_non_oxygen_contacts"] > 0 399 if severe_contact or (OPENMM_PREFLIGHT_ABORT_ON_NA_NON_O_CONTACT and sodium_non_oxygen_contact): --> 400 raise RuntimeError( 401 f"{built.system_name}: periodic contact preflight found severe contacts before OpenMM dynamics: " 402 f"{clash_summary}. Na-O contacts above {OPENMM_PREFLIGHT_ALLOWED_NA_O_CONTACT_A:.2f} A are treated as " 403 "ion coordination rather than clashes."
> 
> RuntimeError: m5: periodic contact preflight found severe contacts before OpenMM dynamics: {'cutoff_a': 2.0, 'raw_pairs': 153473, 'nonexcluded': 1, 'sodium_contacts': 1, 'intramolecular': 0, 'intermolecular': 1, 'min_distance_a': 1.900142955148375, 'severe_contacts': 0, 'sodium_oxygen_contacts': 0, 'sodium_non_oxygen_contacts': 1, 'allowed_sodium_oxygen_contacts': 0, 'details': [{'distance_a': 1.900142955148375, 'sodium_contact': True, 'sodium_oxygen_contact': False, 'intramolecular': False, 'atom_1': 'Na1', 'atom_2': 'Na1', 'residue_1': 'SOD', 'residue_2': 'SOD', 'chain_1': 'A', 'chain_2': 'A', 'residue_number_1': '7103', 'residue_number_2': '7856'}]}. Na-O contacts above 1.80 A are treated as ion coordination rather than clashes.

Sounds like in this case, it didn't resolve the contacts? But note that this case it is the ion colliding with the polymer (or possibly water?).  @joelaforet, is the ion treated as part of the polymer, or is it packed afterwards?

mrshirts:
@joelaforet @timbernat @StephMcCallum @erjank What are the blockers with merging this in at this point?

joelaforet:
I don't think we should claim AA-DPD + ions works yet. That notebook doesn't use the implementation given by this PR. This is the notebook that is the most up-to-date + performs the best: https://github.com/MuPT-hub/mupt-examples/blob/finite-geometry-AA-DPD/examples_system/Finite_Geometry_AA_DPD_PET_Slab_OpenMM.ipynb

erjank:
Agree with Joe about the ions, but otherwise I think this is great to merge. Got the slab notebook working. There are some potential optimizations we can test later, but overall the functionality is there. Great work, Joe.

mupt/builders/all_atom_dpd.py
erjank, line 1: Nice work!

erjank (approved):
LGTM

mrshirts:
Question (this is not blocking for merging but a question): are DPD energy parameters (a/gamma) assigned only to heteroatoms, or to all atoms including H's?  It could be that one could eliminate the use for the H's, since they are usually so close (about 1.0 A) from the heteroatoms that they are contained within the radii.   If possible, that would improve the speed a lot by reducing the number of particles.

mupt/builders/all_atom_dpd.py
StephMcCallum: Just to clarify, bonds, angles, and dihedral parameters are being read in from openff here? Or all they all being set to a default?

StephMcCallum (dismissed):
I think this is good to merge. These comments highlight a future improvement but aren't blocking.

devtools/scripts/validate_all_atom_dpd_pe_melt.py
```diff
@@ -0,0 +1,583 @@
...
+@dataclass(frozen=True)
+class OpenMMDeps:
+    """Optional OpenMM validation imports."""
+
+    Interchange: Any
+    ForceField: Any
+    Molecule: Any
+    Topology: Any
+    ToolkitRegistry: Any
+    NAGLToolkitWrapper: Any
```
timbernat: These are OpenFF, not OpenMM dependencies?

devtools/scripts/validate_all_atom_dpd_pe_melt.py
```diff
@@ -0,0 +1,583 @@
...
+    off_unit: Any
+    omm_unit: Any
+    LangevinMiddleIntegrator: Any
+    MonteCarloBarostat: Any
+    Vec3: Any
+    Simulation: Any
+    primitive_to_rdkit_mols: Any
+
+
+def parse_args() -> argparse.Namespace:
```
timbernat, line 106: Given the CLI-based input of system/sim settings here, is it the case that this script is intended _only_ to be run manually (i.e. not as part of a GH action-based workflow? It may be useful to have this as an automated test whose workflow runs much less frequently, e.g. only runs this script on releases, version tag increments, etc. @dwhswenson likely has a better sense of what to do here

devtools/scripts/validate_all_atom_dpd_pe_melt.py
```diff
@@ -0,0 +1,583 @@
...
+        ToolkitRegistry=ToolkitRegistry,
+        NAGLToolkitWrapper=NAGLToolkitWrapper,
+        off_unit=off_unit,
+        omm_unit=omm_unit,
+        LangevinMiddleIntegrator=LangevinMiddleIntegrator,
+        MonteCarloBarostat=MonteCarloBarostat,
+        Vec3=Vec3,
+        Simulation=Simulation,
+        primitive_to_rdkit_mols=primitive_to_rdkit_mols,
+    )
```
timbernat: Does this need to be a dedicated function, on top of the ImportError a builtin `import ...` statement raises if the target modules aren't installed? We don't want to have all those dependences part of the core stack, which this guards against, but it doesn't prevent them from being required here

devtools/scripts/validate_all_atom_dpd_pe_melt.py
```diff
@@ -0,0 +1,583 @@
...
+    simulation.minimizeEnergy()
+    minimized_energy = energy_kj_mol(simulation, deps.omm_unit)
+    print("OpenMM diagnostics")
+    print(f"  molecule_count: {len(molecules)}")
+    print(f"  atom_count: {n_openmm_atoms}")
+    print(f"  constraints: {system.getNumConstraints()}")
+    print(f"  charge_method: {charge_method}")
+    print(f"  initial_potential_energy_kj_mol: {initial_energy:.6f}")
+    print(f"  minimized_potential_energy_kj_mol: {minimized_energy:.6f}")
+    print(f"  finite_energies: {bool(np.isfinite(initial_energy) and np.isfinite(minimized_energy))}")
```
timbernat: These (and similar elsewhere in this file) out to be `logging` statements

mupt/builders/random_walk.py
```diff
@@ -33,13 +33,19 @@
 from ..mupr.primitives import Primitive, PrimitiveHandle
 
 
+def _random_unit_vector(rng : np.random.Generator, dimension : Dims=3) -> np.ndarray[Shape[Dims], float]:
+    '''Generate a random unit vector from an explicit NumPy generator.'''
+    return normalized(rng.uniform(low=-1.0, high=1.0, size=dimension))
```
timbernat: This is useful for ensuring repeatability of tests! Should live directly in [`geometry.coordinates.directions.random_vector()`](https://github.com/joelaforet/mupt/blob/issue-77-aa-dpd-builder/mupt/geometry/coordinates/directions.py#L12-L23), rather than here; happy to have that function modified to ingest the Optional rng get injected as a kwarg

devtools/scripts/validate_all_atom_dpd_pe_melt.py
timbernat: It's a fair question, looks to me like it'd be better suited for examples, but may be easier to keep close to toolkit code if CI integration is planned. Diffs are relatively small, unlike a notebook, so no overt harm to keeping it in the clearly-marked place it is now

devtools/scripts/validate_all_atom_dpd_pe_melt.py
```diff
@@ -0,0 +1,583 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+import networkx as nx
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+if str(REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(REPO_ROOT))
```
timbernat: This looks a little dicey, is the point just to have mupt dependencies visible without installing them in the local env?

mupt/builders/random_walk.py
```diff
@@ -81,7 +90,7 @@ def random_walk_jointed_chain(
         raise ValueError(f"Random walk starting point must be a {dimension}-dimensional vector")
     
     if initial_direction is None:
-        initial_direction = random_unit_vector(dimension=dimension)
+        initial_direction = random_unit_vector(dimension=dimension) if rng is None else _random_unit_vector(rng, dimension)
```
timbernat: As above, should just add `rng` to random_vector/random_unit_vector and extend the call signatures here

mupt/tests/builders/test_all_atom_dpd.py
timbernat: Probably not, we don't want something that heavy spin up on every CI workflow run on every PR. You have the script that _could_ run what you describe on demand also in this PR, which makes that kind of validation reproducible and doesn't need to be doubled in these tests

mupt/tests/builders/test_all_atom_dpd.py
```diff
@@ -0,0 +1,662 @@
...
+    segment.set_topology(nx.path_graph(handles, create_using=TopologicalStructure), max_registration_iter=100)
+    root.attach_child(segment)
+    return root, resname_map
+
+
+def test_imports_public_symbols_without_hoomd_or_openff(monkeypatch):
+    real_import = builtins.__import__
+
+    def guarded_import(name, *args, **kwargs):
+        if name == "hoomd" or name.startswith("hoomd.") or name == "openff" or name.startswith("openff."):
```
timbernat:
```suggestion
if name.startswith("hoomd") or name.startswith("openff"):
```

`str.startswith()` already covers the exact match

mupt/tests/builders/test_all_atom_dpd.py
```diff
@@ -0,0 +1,662 @@
...
+
+    sys.modules.pop("mupt.builders.all_atom_dpd", None)
+    monkeypatch.setattr(builtins, "__import__", guarded_import)
+
+    module = importlib.import_module("mupt.builders.all_atom_dpd")
+
+    assert module.AllAtomDPDBuilder.__name__ == "AllAtomDPDBuilder"
+    assert module.AllAtomDPDSettings.__name__ == "AllAtomDPDSettings"
+    assert module.AllAtomDPDResult.__name__ == "AllAtomDPDResult"
+    assert module.AllAtomDPDParameterProvider.__name__ == "AllAtomDPDParameterProvider"
```
timbernat: What's the purpose of this name check?

mupt/tests/builders/test_all_atom_dpd.py
```diff
@@ -0,0 +1,662 @@
...
+            return "cpu"
+
+        @staticmethod
+        def GPU():
+            return "gpu"
+
+    class FakeHoomd:
+        device = FakeDevice
+
+    builder = AllAtomDPDBuilder(settings=AllAtomDPDSettings(device=setting))
```
timbernat, line 285: Good mock!

mupt/builders/all_atom_dpd.py
```diff
@@ -0,0 +1,1218 @@
...
+1. Export the updated atom coordinates and periodic box to the target atomistic
+   MD engine.
+2. Build an all-atom force-field system with explicit hydrogens, periodic box
+   vectors, and production-quality partial charges. For OpenFF validation of
+   polyethylene, the NAGL/AshGC model ``openff-gnn-am1bcc-1.0.0.pt`` was used.
+3. Run energy minimization and require finite energies.
+4. Run short NVT dynamics with regular MD settings. For constrained hydrogens,
+   ``2 fs`` and ``1 / ps`` Langevin friction are reasonable smoke-test settings.
+5. Run NPT equilibration at the intended temperature and pressure until density,
+   volume, potential energy, and kinetic energy are bounded and stationary.
```
timbernat, line 29: This really rubs against the "no force field parameters in MuPT" and "we will not be in charge of running simulations" policies we set forth for 1.0; some of these parts might make more sense as utils which accompany an example notebook?

mupt/builders/all_atom_dpd.py
```diff
@@ -0,0 +1,1218 @@
...
+    angle_params: dict[str, dict[str, float]] = field(default_factory=dict)
+    dihedral_params: dict[str, dict[str, float]] = field(default_factory=dict)
+    improper_params: dict[str, dict[str, float]] = field(default_factory=dict)
+    bond_type_by_group: dict[tuple[int, int], str] = field(default_factory=dict)
+    angle_type_by_group: dict[tuple[int, int, int], str] = field(default_factory=dict)
+    dihedral_type_by_group: dict[tuple[int, int, int, int], list[str]] = field(default_factory=dict)
+    improper_type_by_group: dict[tuple[int, int, int, int], list[str]] = field(default_factory=dict)
+    atom_epsilons: dict[int, float] = field(default_factory=dict)
+    atom_types_by_global: dict[int, str] = field(default_factory=dict)
+    epsilon_by_type: dict[str, float] = field(default_factory=dict)
```
timbernat, line 232: As above, do we need to be handling parameter sets for HOOMD directly within the mupt core toolkit?

mupt/builders/all_atom_dpd.py
```diff
@@ -0,0 +1,1218 @@
...
+    def parameterize(
+        self,
+        root: Primitive,
+        records: list[_SegmentRecord],
+        settings: AllAtomDPDSettings,
+    ) -> _ParameterTables:
+        """Label RDKit segment molecules with OpenFF and collect parameters."""
+
+        from openff.toolkit import ForceField, Molecule, Topology
+        from openff.units import unit
```
timbernat, line 294: (discussion) We _really_ don't want to have openff as a base toolkit dependency; that's guarded in this invocation, but not by much. This implicit import is "good enough for now", those kinds of solutions tend to hang around long past when we'd like them to. Worth discussing if it's possible to move openff-based parts outside before 1.0

mupt/builders/all_atom_dpd.py
```diff
@@ -0,0 +1,1218 @@
...
+                i = queue.popleft()
+                for j in neighbors[i]:
+                    if j in seen:
+                        continue
+                    delta = positions[j] - positions[i]
+                    delta -= box_lengths * np.round(delta / box_lengths)
+                    unwrapped[j] = unwrapped[i] + delta
+                    seen.add(j)
+                    queue.append(j)
+        return unwrapped
```
timbernat, line 1517: Not saying it need to be done here, but we'll need a more generic, topology- and shape-aware position unwrapping algorithm by release. Happy to head this up in future

mupt/builders/all_atom_dpd.py
```diff
@@ -0,0 +1,1218 @@
...
+
+
+class AllAtomDPDBuilder:
+    """All-atom DPD builder for SAAMR Primitive hierarchies."""
+
+    def __init__(
+        self,
+        settings: Optional[AllAtomDPDSettings] = None,
+        parameter_provider: Optional[AllAtomDPDParameterProvider] = None,
+        placement_generator_factory: Callable[[np.random.Generator, float], PlacementGenerator] | None = None,
```
timbernat: Why is this passed as a factory, rather than an initialized PlacementGenerator object? The default could receive the output of `self._default_placement_generator()`, rather than the method itself. This has the veneer of generality, but is set up in a way pretty specific to how the default's config is handled internally

timbernat (changes requested):
Left some feedback, in summary:
* Tests look good; more expensive DPD sim + energy check should remain in validation script, rather than being moved to unit tests
* RNG input can be put directly into `geometry.coordinates.directions.random_vector()` and calls referencing it updated
* Lots of heavy dependencies, especially OpenFF _with_ parameterization toolkits introduced; will make managing installs and deps much trickier

My biggest concern is that much of the content here goes squarely against the "mupt will not handle force field parameterization" and "mupt will not run simulations for you" constraints we had set out early on. It's worth making sure we're okay with architechturally, as "temporary" solutions rarely stay temporary

mupt/builders/all_atom_dpd.py
joelaforet: Edited to take an initialized PlacementGenerator.

mupt/builders/all_atom_dpd.py
joelaforet: We can revisit this in subsequent PR's, I think this is fine for a MVP though.

mupt/builders/all_atom_dpd.py
joelaforet: For AA-DPD I think so. Unless we decide to move the implementation of this to FlowerMD or PhantomWalk and then import those routines in the mupt core.

mupt/builders/all_atom_dpd.py
joelaforet: This is just a docstring with suggestions. MuPT isn't doing these steps in the toolkit, it's guidance for users after they run the DPD initializer.

mupt/tests/builders/test_all_atom_dpd.py
joelaforet: Removed!

mupt/tests/builders/test_all_atom_dpd.py
joelaforet: Fixed!

mupt/builders/random_walk.py
joelaforet: Done!

devtools/scripts/validate_all_atom_dpd_pe_melt.py
joelaforet: Yes, as a safety net in case someone didn't run "pip install -e .". I removed this and opted to tell people to do that via a docstring.

mupt/builders/random_walk.py
joelaforet: Done! And modified the geometry module + tests accordingly.

devtools/scripts/validate_all_atom_dpd_pe_melt.py
joelaforet: Logged!

devtools/scripts/validate_all_atom_dpd_pe_melt.py
joelaforet: Function is removed.

devtools/scripts/validate_all_atom_dpd_pe_melt.py
joelaforet: Oops!

mupt/builders/all_atom_dpd.py
erjank, line 1: Note that for small boxes we can use a Tree neighbor list to get around the cell list limitations on 3x r_cut

erjank (dismissed):

erjank (approved):
LGTM

erjank:
I think the openff dependencies are OK, at least for now, since they provide necessary functionality. It may make sense to move them to mupt-examples when we make the AA-DPD initializer more agnostic to the forcefields passed in.

erjank:
Looks like the RNG requested changes were made. Since waiting on this merge is blocking the CG initialization testing we would like to do, I am going to beg forgiveness later and merge this in. Nice work all!

PR #94, opened 2026-08-06
Add `to_mbuild` and `to_gmso` methods to `mupt/interfaces/mosdef`

### LINKED ISSUE #25: Support conversion between mBuild and MuPT representation
The MoSDeF ecosystem, and in particular [GMSO](https://gmso.mosdef.org/en/stable/), provide flexible representations of chemical systems *with* force fields parameters which would enable many downstream capabilities including:
* Typing and parameterization of CG systems
* Export to CG-supporting molecular dynamics engines such as HOOMD and LAMMPS
* Injection of custom force field functional forms (e.g. FENE bonds, DPD potentials)
* Bespoke parameterization for non-atomic systems, with parameters supplied by user

However, using these tools with MuPT-represented systems requires an entry point into the MoSDeF ecosystem.

Converters from MuPT Primitives to [mBuild](https://mbuild.mosdef.org/en/stable/) Compounds, namely
* An exporter (MuPT -> mBuild)
* An importer (mBuild -> MuPT)
* Preservation of hierarchical information in either conversion direction
* Preservation of Primitive (node) labels, to ensure parameter inference from types downstream in GMSO is undisturbed

Native tooling already exists to [convert mBuild Compounds into GMSO objects](https://github.com/mosdef-hub/mbuild/blob/main/mbuild/conversion.py#L1861-L1891), so no engineering would be required for downstream functionality within MuPT.

For all-atom systems only, mBuild supplies [from_rdkit](https://github.com/mosdef-hub/mbuild/blob/main/mbuild/conversion.py#L805-L874)/[to_rdkit](https://github.com/mosdef-hub/mbuild/blob/main/mbuild/conversion.py#L1715) converters. 

**Additional context**
There are several reasons that MuPT <-> mBuild in particular (as opposed to adjacent tools) is 
* The internal molecular representation of mBuild compounds and MuPT Primitives correspond closely and very naturally
  * Both formats distinguish atomic units by injecting element info, so AA workflow are subsumed by a general converter as long as that info is handed off
* RDKit imposes a flat molecular representation which precludes preservation of the arbitrary intermediate levels of organization that Primitives or Compounds encode 
* A direct to GMSO pipeline is less natural, because MuPT representations do not (and will not) carry around force field parameters, and because GMSO also requires a flat graph of structural units similar to RDKit

### LINKED ISSUE #28: GMSO system interface
Need: write functions and tutorial notebooks demonstrating how to go from MuPT repr -> mbuild -> GMSO -> LAMMPS/HOOMD simulation. The MuPT repr -> mbuild will be covered in issue #25.

Purpose: Enable CG and AA simulations from MuPT representation.

chrisjonesBSU:
After our conversation last week, one of the to-dos was to help provide a useable interface to go from MuPT to MoSDeF. This PR adds this for converting to mbuild and gmso. If this gets merged, then we can start looking at adding MoSDeF to MuPT support, and the more interesting and complex mupt/mosdef hand-offs that might be possible (back-mapping, coarse-graining, coordinate generation, etc.)

The primary design and approach is as follows:

1. To convert to an mBuild Compound (or GMSO topology), the user needs to define what the bottom resolution is that they want to convert (bonds in mosdef only exist at the bottom). 
2. The goal here is to allow flexibility in making that choice, instead of always assuming that the bottom is where nodes are explicitly labelled as atoms or leafs.
3. This should all be possible without requiring that the user changes or mutates anything in their Primitive object.
4. We should retain as much information as possible about the hierarchy **above** the chosen floor: This is possible in mBuild to an arbitrary degree of higher resolutions, and also means that information like `Molecule` and `Residue` in gmso `Topology` is retained consistently from MuPT. The information that is retained at these higher-levels of resolution includes labels (i.e., names), and consistent grouping of children (i.e., particles). Bonds/edges at this level aren't carried into MoSDeF (as mentioned before, bonds are always at the bottom of the hierarchy in MoSDeF)

Flexible resolution slicing/stopping is achieved by the following additions to `interfaces/_shared/topology.py`

- `ResolutionSpec`: This provides multiple options for users to define **what** a bottom level of a Primitive is.
- `ResolutionPredicate` A callable built from a `ResolutionSpec` that returns a `bool`. This follows the signature/design of the `stop` parameter in [anytree's iterators](https://anytree.readthedocs.io/en/stable/api/anytree.iterators.html)
- `_resolve_to_resolution`: Inspired by the existing `_resolve_to_atom`, but generalized to accept other levels (as opposed to always atomistic). This copies the same check/validation structure as `_resolve_to_atom` which I believe guards against broken/inconsistent topologies.
- `resolution_graph`: Builds a flat graph, that contains a single "slice" of the Primitive (the slice being defined by the items above). This potentially has multiple use-cases, but for this PR, it serves as the template that defines particles/bonds in MoSDeF.

All the changes exist outside the core of MuPT. The new methods are placed in`interfaces/_shared/topology.py` as they should be useful for other interfaces, or writers that MuPT builds internally.

[Here is a Gist that shows the workflow](https://gist.github.com/chrisjonesBSU/b1b42bf8637b7a6e6d3bf89a9f8fe5db) if you install MuPT from my fork.

### Changes I made:
1. I think a single `interfaces/mosdef` directory makes more sense than one of (or both) `interfaces/gmso` and `interfaces/mbuild`. So, I made a `mosdef` directory in `interfaces`, moved all the files from `interfaces/gmso` into it, and that is now where `exporters.py` lives.
2. This PR adds a `to_gmso` and `to_mbuild`, but `to_gmso` is a short wrapper that calls `to_mbuild`. I think this is much preferable over writing a gmso-specifc `to_gmso` method as it leverages already existing conversions built into GMSO (from_mbuild). The user never sees the intermediate `to_compound` step when calling `to_gmso`, and only gets the `gmso.core.Topology`. So, the use-cases are clearly delineated from the user-perspective.

### Changes to think about (I left as #DEVNOTE in the files changed)
1. Some of the workflows added here would have benefited from something along the lines of `Primitive.sibling_connections` that mirrors `Primitive.external_connections` and `Primitive.internal_connections`. I'm not totally sure how you would implement that as of now (still wrapping my head around MuPT/Primitive), but it could be a nice addition for users to call directly, or make future development choices easier.
2. Maybe `_resolve_to_resolution` can replace `_resolve_to_atom` as it should work the same given the correct `ResolutionPredicate`. One approach is that _resolve_to_atom could be a wrapper that builds the right predicate and calls `resolve_to_resolution`. This should reduce some duplicated code/efforts in `interfaces/_shared/topology.py` 

Unit tests are still needed (pending review and discussion around the changes)

Let me know your thoughts on the approach.

**AI Disclosure:** I used Claude to help debug errors when testing these additions as I was building them. This involved Claude making edits to the code directly. I carefully reviewed and vetted any edits/additions made by AI.


chrisjonesBSU:
Also, many of the in-line code comments were notes for myself and to help with review and discussion, I can go through and remove or trim down on them.

PR #100, opened 2026-09-10
New docs skeleton

dwhswenson:
## Description
This updates the skeleton of the docs to be a little more modern and useful, and to start to include actual content from MuPT. 

## Todos
  - [x] Refactor documentation according to [Diátaxis](https://diataxis.fr) principles.
  - [x] Add automated API doc generation
  - [x] Add tools to download notebooks from must-examples that can be re-used in documentation
  - [x] Update RTD setup
  - [x] Add placeholders for docs that others will need to write

## Status
- [x] Ready to go

There's plenty of room within what's added to improve appearance (e.g., layout of icons on the first page; possibly the layout of the API docs). The goal here is to get a first usable version in place. Similarly, the current how-to/tutorial split is not correct (both are probably how-tos), but organizing those is for a future PR.


docs/conf.py
```diff
@@ -32,6 +35,30 @@
 # The full version, including alpha/beta/rc tags
 release = ''
 
+examples_repo_base = (
+    'https://raw.githubusercontent.com/MuPT-hub/mupt-examples/main/'
+)
```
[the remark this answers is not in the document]
dwhswenson: The intent is for the documentation here to evolve as the examples evolve. Pinning is counterproductive.

dwhswenson:
"Critical" issue identified by Copilot was that I removed the placeholder code (presumably from the cookie cutter) that printed out a mangled Thoreau quote. I don't think any downstream users were depending on that functionality.

It also complains that docs/README is out of date now; might be worth just removing that file (also looks like a cookie cutter thing).

CI failing seems to be legit problems outside the changes in this PR. Holding here until we can get CI to green.

dwhswenson:
CI passing; ready for review @timbernat! Feel free to review the diffs, but for docs, also take a look at the RTD bot's comment, and try the "Preview build" button.

There's still some work to make a few things look better, but the goal of this PR is to put things in place so that we can start to fill in the content.

.gitignore
```diff
@@ -122,3 +125,6 @@ ENV/
 */_version.py
 # Cookiecutter-CMS Test Artifacts
 /tests/CI_files
+# pixi environments
```
timbernat, line 128: (praise) Good to have early, though there's no pixi config yet

docs/conf.py
```diff
@@ -51,15 +79,72 @@
...
+    'members',
+    'undoc-members',
+    'show-inheritance',
+    'show-module-summary',
+]
+
+# Parse internal modules so AutoAPI can resolve public annotations and base
+# classes, but do not render them. New modules inside the supported packages
+# are discovered automatically; promoting a top-level package is explicit.
+public_api_prefixes = (
```
timbernat, line 111: (clarification, non-blocking) What's the rationale behind these API selections, is it just because they're top-level? E.g. roles is only toplevel to avoid long relative imports, mupt.mupr should probably have it's internals referenced explicitly (.connections, .primitives, etc).

This can all be updated later as docs are populated, just wanted to flag so I understood the organization

docs/getting_started/index.rst
```diff
@@ -0,0 +1,7 @@
+Getting Started
```
timbernat, line 1: (TODO) Will update as docs and example notebooks evolve

docs/conf.py
timbernat: Also, presumably the autodoc extension will update the sub-pages as submodules under these public modules as they change (e.g. mupr.connection breaking into sup-parts following #82), is that right?

timbernat (approved):
Left som enon-blocking clarification questions on API headers in docs, but skeleton looks good! Will fill in sections as part of #101 - #108 in subsequent PRs

